PR #12, opened 2016-07-13
Match API for response status/proto handling

p0pr0ck5:
libmodsecurity now handles response status and HTTP protocol via
msc_process_response_headers().

I know @phantom-az is working on this is his branch, but I think this should be on master now, because right now builds are currently broken.


zimmerle:
Merged already ;)

PR #20, opened 2016-11-09
Fix duplicate rule IDs in tests

defanator:


zimmerle:
Thanks @defanator. Merged.

PR #24, opened 2016-12-01
Tests: add test for response body

defanator:
Blocking requests basing on contents of  the response body is not working at the moment, so this test is marked as TODO.

This is what happens with the configuration from this test:

    $ curl -vvi http://127.0.0.1:8873/body1
    * Hostname was NOT found in DNS cache
    *   Trying 127.0.0.1...
    * Connected to 127.0.0.1 (127.0.0.1) port 8873 (#0)
    > GET /body1 HTTP/1.1
    > User-Agent: curl/7.38.0
    > Host: 127.0.0.1:8873
    > Accept: */*
    > 
    * Empty reply from server
    * Connection #0 to host 127.0.0.1 left intact
    curl: (52) Empty reply from server

### DIFF
--- a/tests/modsecurity-response-body.t
+++ b/tests/modsecurity-response-body.t
@@ -1,0 +1,69 @@
+#!/usr/bin/perl
+
+# (C) Andrei Belov
+
+# Tests for ModSecurity-nginx connector (response body operations).
+
+###############################################################################
+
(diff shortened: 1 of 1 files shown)


zimmerle:
Merged ;) thanks.

PR #28, opened 2016-12-06
Extend PCRE malloc/free wrappers to use nginx pools

defanator:
The goal is to completely avoid direct malloc()/free() calls from PCRE.

Implementation is heavily based on corresponding parts of the Lua nginx module:
https://github.com/openresty/lua-nginx-module/blob/master/src/ngx_http_lua_pcrefix.c


zimmerle:
Merged thanks!

PR #49, opened 2017-06-14
tests: update request body test with more RFC2616 methods

defanator:


zimmerle:
Merged ;) thanks! :+1:

PR #64, opened 2017-08-23
Update to latest nginx versions and fix tests

defanator:


zimmerle:
Hi @defanator,

Thank you very much ;) It is all merged.

PR #78, opened 2017-12-18
Fix header already sent

dennus:
Complete response body filtering

### DIFF
--- a/src/ngx_http_modsecurity_pre_access.c
+++ b/src/ngx_http_modsecurity_pre_access.c
@@ -180,17 +180,12 @@ ngx_http_modsecurity_pre_access_handler(ngx_http_request_t *r)
              * it may ask for a intervention in consequence of that.
              *
              */
-          
-        }
-        /** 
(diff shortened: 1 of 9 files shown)


src/ngx_http_modsecurity_pre_access.c
```diff
@@ -180,12 +180,17 @@ ngx_http_modsecurity_pre_access_handler(ngx_http_request_t *r)
              * it may ask for a intervention in consequence of that.
              *
              */
-            ret = ngx_http_modsecurity_process_intervention(ctx->modsec_transaction, r);
+          
+        }
+        /** 
+         *  @dennus: I'm move transaction outside cycle, because I'm think that was uselless calls
```
zimmerle: Not really. The rules can set limit for the size of the body, for example. So, whenever the size is over certain threshold modsecurity takes an action. Having that check when the body is full already, it may be  too late for the rule take an action.

dennus:
@zimmerle, thanks for comment

PR #84, opened 2017-12-20
Fix response body

dennus:
Complete response processing, I'm changed two tests in modsecurity.t, because now response validation may bring to errors like 403, 401, etc. But Nginx special responses with code < 400 returned only on phase 3

### DIFF
--- a/src/ngx_http_modsecurity_body_filter.c
+++ b/src/ngx_http_modsecurity_body_filter.c
@@ -201,9 +201,9 @@ ngx_http_modsecurity_body_filter(ngx_http_request_t *r, ngx_chain_t *in)
         ngx_http_modsecurity_pcre_malloc_done(old_pool);
         ret = ngx_http_modsecurity_process_intervention(ctx->modsec_transaction, r);
         if (ret > 0) {
-            if (ret < NGX_HTTP_BAD_REQUEST && ctx->header_pt != NULL)
+             if (ret < NGX_HTTP_BAD_REQUEST && ctx->header_pt != NULL)
                 ctx->header_pt(r);
-            return ngx_http_filter_finalize_request(r,
+             else return ngx_http_filter_finalize_request(r,
                     &ngx_http_modsecurity_module, ret);
         } else if (ret < 0) {
             return ngx_http_filter_finalize_request(r,
--- a/src/ngx_http_modsecurity_body_filter.c
+++ b/src/ngx_http_modsecurity_body_filter.c
@@ -34,7 +34,6 @@ ngx_http_modsecurity_body_filter_init(void)
 ngx_int_t
 ngx_http_modsecurity_body_filter(ngx_http_request_t *r, ngx_chain_t *in)
 {
-
     ngx_http_modsecurity_ctx_t *ctx = NULL;
     ngx_chain_t *chain = in;
     ngx_int_t ret;
@@ -47,15 +46,17 @@ ngx_http_modsecurity_body_filter(ngx_http_request_t *r, ngx_chain_t *in)
     ngx_uint_t i = 0;
 #endif
 
-    if (in == NULL) {
+if (in == NULL) {
+         ngx_log_debug0(NGX_LOG_DEBUG_HTTP, r->connection->log, 0, "MDS input chain is null");
+        
         return ngx_http_next_body_filter(r, in);
-    }
+  }
 
     /* get context for request */
     ctx = ngx_http_get_module_ctx(r, ngx_http_modsecurity_module);
     dd("body filter, recovering ctx: %p", ctx);
 										   
-    if (ctx == NULL || r->filter_finalize) {
+    if (ctx == NULL || r->filter_finalize || ctx->response_body_filtered) {
         return ngx_http_next_body_filter(r, in);
     }
 
@@ -143,18 +144,16 @@ ngx_http_modsecurity_body_filter(ngx_http_request_t *r, ngx_chain_t *in)
         ngx_buf_t *copy_buf;
         ngx_chain_t* copy_chain;
         is_request_processed = chain->buf->last_buf;
-        ngx_int_t data_size = chain->buf->end - chain->buf->start;
-        ngx_int_t data_offset = chain->buf->pos - chain->buf->start;
-        u_char *data = chain->buf->start;
+        ngx_int_t data_size = chain->buf->last - chain->buf->pos;
+        u_char *data = chain->buf->pos;
         msc_append_response_body(ctx->modsec_transaction, data,
-                chain->buf->end - data);
+                chain->buf->last - data);
         ret = ngx_http_modsecurity_process_intervention(ctx->modsec_transaction,
                 r);
         if (ret > 0) {
             return ngx_http_filter_finalize_request(r,
                     &ngx_http_modsecurity_module, ret);
         }
-        if (data_size > 0){
             copy_chain = ngx_alloc_chain_link(r->pool);
             if (copy_chain == NULL) {
                 return NGX_ERROR;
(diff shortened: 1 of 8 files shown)


src/ngx_http_modsecurity_body_filter.c
```diff
@@ -130,56 +133,87 @@ ngx_http_modsecurity_body_filter(ngx_http_request_t *r, ngx_chain_t *in)
...
+    for (chain = in; chain != NULL; chain = chain->next) {
+
+        ngx_buf_t *copy_buf;
+        ngx_chain_t* copy_chain;
+        is_request_processed = chain->buf->last_buf;
+        ngx_int_t data_size = chain->buf->end - chain->buf->start;
+        ngx_int_t data_offset = chain->buf->pos - chain->buf->start;
+        u_char *data = chain->buf->start;
+        msc_append_response_body(ctx->modsec_transaction, data,
+                chain->buf->end - data);
```
turchanov: This is wrong. Actual data must be referenced by .pos and .last. See #104

dennus:
@turchanov 

Thanks for you comment, now I'm fix buffer processing.

Unfortunely, now we are not have API to get value of directive SecResponseBodyAccess in connector

defanator:
@zimmerle valid concern about accessing various things like SecResponseBodyAccess from connectors. Is it something that could be achieved in the library on the cheap?

zimmerle:
That is indeed a valid concern. There are a few circumstances where the body is not considered or it is partially considered:

 - Body is bigger than a given threshold (set by the user). This can lead to a partial analysis, or reject (403), again, up to the user.
 - The content of the body is not interesting to be inspected. Choose by the user given the content type header.
 - The content body inspection is complete disabled.

All those should checks are treated by the appendResponseBody. It just proceed with a copy if that is really necessary.  Since the content is already a buffer (in some format), just a pointer is stacked. So, I don't expect to have a massive performance impact. 

I would avoid any decision like that to be taken under the connector side, unless it is extremely necessary. By design we want the connector to be as simple as possible, with no logic, simple bridging the content towards the library. With that, we expect to have easy integration with 3rd party software.

Let me check the lib code to understand if it is doing what it meant to be doing and get back to you.

zimmerle:
Content complete disable is not being checked by _appendResponseBody_. Making a patch and get back to you with a new branch for testing.

zimmerle:
@turchanov, @defanator, @dennus do you mind to test it again? make sure to include the commit: SpiderLabs/ModSecurity@42a472adbda21e6ecc4711fd37d704cdb3d98fbb. It is now considering if the response body was enabled or disabled.

dennus:
@zimmerle, yes, I'm it tested it again. All ok, but unfortunely we still can't have a way to not starting processing response body, because descision makes in processResponseBody and we still must copyng all chunks and create chain copy.

zimmerle:
At first sight I was concerned about a copy of the data to be duplicated inside ModSecurity. 
Reading your code carefully, I saw the duplication of the chain. Correct me if I am wrong, but it seems like you have to hold the request/response till it is complete in order to processing it with ModSecurity, after that, deliver it to the next module in chain.

Question: can we delivery the content to ModSecurity by pieces? without duplicate the chain?

I am thinking in a good way to extract the information of a configuration-location in a good manner. Don't want to make it mandatory for the connector implementation, but, useful for optimizations.

@dennus @defanator

dennus:
We call msc_append_response_body every time when receive next chunk of response body. But yes, whe hold response body in chain copy, because if we don't do this, nginx write filter will send to client body chunks before sending headers (which we hold before in header filter). If we don't hold headers, we cannot change any of them (for e.g., we cannot change response state code) because after it sending we get "header already sent" error.

zimmerle:
I've got your point, and it seems like the major performance problem is duo to this chain hold. That leads to a question: isn't this chain hold also necessary in order to have proper headers set in inspections on phases grater than 3? Regardless being inspecting the body or not.

dennus:
As far as I can see, this true - any modification on headers after it's sending to client throws error, that if we are trying to modify it after phase 3, we must hold a chain (except a case on 4 phase,  when first call in msc_process_intervention in chunks processing cycle returns negative result). Another words, if we want to send modified headers, we must do this before sending a body). Maybe @defanator correct me.

From another point a view, does msc_process_intervention may return non zero result before or after msc_process_response_body on phase 4, if SecResponseBodyAccess turned to off? If not, we may not to hold response body chain, because headers will never be modify.

zimmerle:
Can you double-check if that is still an issue after merging #165?

mtx-z:
Hello,

I don't know exactly why, but it seems that we're having the same issue for a few weeks. 
The build is from last month (latest nginx and modsec from last month). And I use OWASP core rules, with a classic config. 

Modsecurity correctly catches and deny the request. But I fear that requests are still passed to the PHP upstream socket, as I have "500 error headers already sent". 

I wanted to know if it can be related to a (mis)configuration I made, as I'm unable to find the issue...

Below are the 2 log lines: first is the catch, second is the 500. All Modsec catch will throw a Header already sent 500 after. 
The user never gets the HTTP deny response, but a "socket disconnected" error. 

Thing is, the ModSec logline contains parts of the 500 error text "while sending to client,".

In the end the request is catched, logged, and so fail2ban can detect it. But I fear that the request is still passed to the PHP socket: as I have the headers already sent. 
Also, the modsecurity debug log contains the full HTML body response. 

Thanks

```
2021/11/05 23:59:31 [error] 8089#8089: *1332 [client 88.164.220.225] ModSecurity: Access denied with code 403 (phase 4). Matched "Operator `Ge' with parameter `4' against variable `TX:OUTBOUND_ANOMALY_SCORE' (Value: `4' ) [file "/etc/nginx/modsec/base_rules/RESPONSE-959-BLOCKING-EVALUATION.conf"] [line "129"] [id "959100"] [rev ""] [msg "Outbound Anomaly Score Exceeded (Total Score: 4)"] [data ""] [severity "0"] [ver "OWASP_CRS/3.4.0-dev"] [maturity "0"] [accuracy "0"] [tag "anomaly-evaluation"] [hostname "51.XXX.XXX.XXX"] [uri "/"] [unique_id "1636153170"] [ref ""] while sending to client, client: 88..XXX.XXX.XXX, server: www.domain.com, request: "GET /?s=%22%2Bresponse.write%28%5B764%C2%A0873*728%C2%A0920%29%2B%22 HTTP/1.1", upstream: "fastcgi://unix:/var/run/php7-fpm.sock", host: "domain.com"
2021/11/05 23:59:31 [alert] 8089#8089: *1332 header already sent while sending to client, client: 88.XXX.XXX.XXX, server: www.domain.com, request: "GET /?s=%22%2Bresponse.write%28%5B764%C2%A0873*728%C2%A0920%29%2B%22 HTTP/1.1", upstream: "fastcgi://unix:/var/run/php7-fpm.sock", host: "domain.com"
```

airween:
What is your [SecRequestBodyAccess](https://github.com/SpiderLabs/ModSecurity/wiki/Reference-Manual-(v2.x)#secrequestbodyaccess) value?

mtx-z:
@airween the value is `on`: `SecRequestBodyAccess On`. I also have 
```
SecRequestBodyLimit 100000000
SecRequestBodyNoFilesLimit 131072
```

Dr-Lazarus-V2:
Apologies, I believe there is another dimension to this issue.

I have observed that when the size of the response body crosses a particular threshold, even if the ModSecurity engine detects and throws a 403, the client still receives the complete request with a status of 200.

For example, if I add a custom regex rule to detect the word "jolly" in the response, the client receives the response with a status 200, even though it is logged as 403 in the audit logs and 500 in the Nginx logs (due to headers already being sent).

PR #87, opened 2017-12-22
The pool pointer is now available for ngx_http_modsecurity_config_cleanup

AirisX:
I did a little research and found out that the rules that are initialized in the pool by the `msc_rules_add_file` or `msc_rules_add` can not be freed by the `msc_rules_cleanup` because `ngx_http_modsecurity_config_cleanup` does not know anything about this pool (since it is NULL).

### DIFF
--- a/src/ngx_http_modsecurity_module.c
+++ b/src/ngx_http_modsecurity_module.c
@@ -554,6 +554,7 @@ static void *ngx_http_modsecurity_create_conf(ngx_conf_t *cf)
     conf->sanity_checks_enabled = NGX_CONF_UNSET;
     conf->rules_set = msc_create_rules_set();
     conf->modsec = NULL;
+    conf->pool = cf->pool;
 
     cln = ngx_pool_cleanup_add(cf->pool, 0);
     if (cln == NULL) {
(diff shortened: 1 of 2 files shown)


AirisX:
It seems that this PR is also reasonable for https://github.com/SpiderLabs/ModSecurity/issues/1546

src/ngx_http_modsecurity_module.c
```diff
@@ -649,7 +650,7 @@ ngx_http_modsecurity_config_cleanup(void *data)
 
     dd("deleting a loc conf -- RuleSet is: \"%p\"", t->rules_set);
 
-    old_pool = ngx_http_modsecurity_pcre_malloc_init(NULL);
+    old_pool = ngx_http_modsecurity_pcre_malloc_init(t->pool);
```
zimmerle: Regardless the parameter here I guess the end result will be the same.

zimmerle: oh wait, i think i've get your point.

AirisX:
@defanator @zimmerle I did a rebase to apply patch more cleanly.

zimmerle:
Merged! Thanks!

PR #105, opened 2018-04-19
Fixed processing of response body chunks ...

turchanov:
in ngx_http_modsecurity_body_filter.

A body filter function (ngx_http_modsecurity_body_filter in our case) can be
called by Nginx several times during request processing. And each time with
it own unique set of chained buf pointers.

For example, suppose a complete response consists of this chain of data:
    A->B->C->D->E
Ngix may (and actually does, as verified by me in gdb) call body filter two
times like this:
    handler(r, in = A->B->C)
    handler(r, in = D->E), E has last_buf set

Current implementation delays feeding chain->buf to msc_append_response_body
until it comes upon a chain with buf->last_buf set. So we loose chain containing
A->B->C sequence. We must process body bufs as soon as we see them in body
handler otherwise we will not see them again.

N.B. You have PR #84 pending. It goes further and fixes the problem when
a blocking decision is made after headers were sent. I intentionally retained
current (buggy) behavior to make my patch less intrusive and easier to review.
Besides #84 impose an excessive memory usage due to a complete copy of all
bufs passed through body filter (we have sometimes 500K and more replies in our
applications) - I will elaborate on this in code review for #84.

This PR depends on #104

### DIFF
--- a/src/ngx_http_modsecurity_body_filter.c
+++ b/src/ngx_http_modsecurity_body_filter.c
@@ -35,7 +35,6 @@ ngx_http_modsecurity_body_filter_init(void)
 ngx_int_t
 ngx_http_modsecurity_body_filter(ngx_http_request_t *r, ngx_chain_t *in)
 {
-    int buffer_fully_loadead = 0;
     ngx_chain_t *chain = in;
     ngx_http_modsecurity_ctx_t *ctx = NULL;
 #if defined(MODSECURITY_SANITY_CHECKS) && (MODSECURITY_SANITY_CHECKS)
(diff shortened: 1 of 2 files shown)


src/ngx_http_modsecurity_body_filter.c
```diff
@@ -135,47 +134,43 @@ ngx_http_modsecurity_body_filter(ngx_http_request_t *r, ngx_chain_t *in)
...
+    int is_request_processed = 0;
     for (; chain != NULL; chain = chain->next)
     {
-/* XXX: chain->buf->last_buf || chain->buf->last_in_chain */
-        if (chain->buf->last_buf) {
-            buffer_fully_loadead = 1;
+        u_char *data = chain->buf->pos;
+        int ret;
+
+        msc_append_response_body(ctx->modsec_transaction, data, chain->buf->last - data);
```
defanator: Let's use `ngx_buf_size()` here.

src/ngx_http_modsecurity_pre_access.c
```diff
@@ -163,10 +163,10 @@ ngx_http_modsecurity_pre_access_handler(ngx_http_request_t *r)
 
         while (chain && !already_inspected)
         {
-            u_char *data = chain->buf->start;
+            u_char *data = chain->buf->pos;
 
             msc_append_request_body(ctx->modsec_transaction, data,
```
defanator: Ditto (it seems like the commit from #104 though, I left the same comment there).

src/ngx_http_modsecurity_body_filter.c
turchanov: `ngx_buf_size` is defined as
```
#define ngx_buf_size(b)                                                      \
    (ngx_buf_in_memory(b) ? (off_t) (b->last - b->pos):                      \
                            (b->file_last - b->file_pos))

```
If somehow we happend to receive a `buf` which is completely in a file (i.e., buf->in_file == true but ngx_buf_in_memory(buf) == false) then .pos = .last = 0 but .file_last - .file_post > 0, so we will try to dereference a non-zero length block at NULL pointer .pos. But with current `.last - .pos`  we would erroneously assume that we have no data but won't segfault at least.

But from the other hand I did a quick testing with gdb, even when a response body is buffered to a temporary file, nginx supplies a `buf` which has both `.in_file` and `.temporary` set (`ngx_buf_in_memory(buf) == true` =>  `.pos`, `.last` are valid).
ngx_http_charset_filter_module also references .pos/.last in `ngx_http_charset_recode_{to,from}_utf8` without worries . So I think it is guaranteed that response body filters get chain bufs with `ngx_buf_in_memory(buf) == true` ?

One more thing, I did a quick grep of Nginx sources and see that ngx_buf_size is not used that much, whereas `.last - .pos` is much more common... maybe we should stick to the latter?

defanator: @turchanov well ok - let's keep `.last - .pos`.

zimmerle:
Merged. Thanks ;)

PR #116, opened 2018-06-25
Increase log level for disruptive actions to error

victorhora:
This pull request is intended to change the log level from _warn_ to the "default" of _error_ when writing log entries of disruptive actions on Nginx [error_log](https://nginx.org/en/docs/ngx_core_module.html#error_log).

I've noticed this would be contradictory to @defanator's commit https://github.com/SpiderLabs/ModSecurity-nginx/commit/b51e555d39337637f8be8c5e16dbfb28b385088a which although it seems technically accurate it seems like it was getting some users a bit confused of why they were not getting ModSecurity logs on Nginx by default. This is intended to resolve issue https://github.com/SpiderLabs/ModSecurity-nginx/issues/112

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -1,5 +1,7 @@
 v1.0.x - YYYY-MMM-DD (To be released)
 -------------------------------------
+ - Increase log level for disruptive actions to "error"
+   [Issue #112 - @victorhora ]
  - Fix processing of response body when gzip compression is enabled
    [Issue #107 - @turchanov]
  - Fixed processing of response body chunks in
--- a/src/ngx_http_modsecurity_module.c
+++ b/src/ngx_http_modsecurity_module.c
@@ -151,7 +151,7 @@ ngx_http_modsecurity_process_intervention (Transaction *transaction, ngx_http_re
         log = "(no log message was specified)";
     }
 
-    ngx_log_error(NGX_LOG_WARN, (ngx_log_t *)r->connection->log, 0, "%s", log);
+    ngx_log_error(NGX_LOG_ERR, (ngx_log_t *)r->connection->log, 0, "%s", log);
 
     if (intervention.log != NULL) {
         free(intervention.log);


zimmerle:
Merged.

robinkb:
@zimmerle Can this change be re-evaluated? It seems strange to me to log disruptive actions through the NGINX error log while ModSecurity also has an audit logger that grants much more control over what is logged, where the logs go, and in what format.

victorhora:
One of the goals with v3 is also to make the behaviour not too different from v2 and this changeset was also meant to provide this similar experience with v2. But we should consider making this a configuration flag so as to allow the flexibility for different use cases. What do you think?

robinkb:
A configuration flag would be very useful. At the moment I am forced to carry a patch to revert the log level to `warn`. Making this configurable would be a lot better.

Given a clean slate without considering current behavior, I would have the ModSecurity connector log disruptive actions in NGINX's error log at `info` level. After all, I expect ModSecurity to block transactions, so if it does, I don't consider it an error. If I need information about specific transactions, I can use ModSecurity's audit logger, where I have a lot more control over the content and format of the logger.

I understand the reasoning that this might be confusing for new users, who just want to verify that ModSecurity is working at all. But might it not be better to document some recommended configuration for debugging purposes? For example, new users could temporarily raise the NGINX's error log level to `warn` (or whichever level is relevant), to quickly verify that ModSecurity is working. Another option would be to give users a quick explanation about the ModSecurity audit logger, or point them to the right documentation.

mtorromeo:
I had to raise the error_log level to info to see some destructive actions in the logs. The issue was already reported https://github.com/SpiderLabs/ModSecurity-nginx/issues/112#issuecomment-403084693

```
[info] 942109#0: *101129 ModSecurity: Warning. Matched "Operator `PmFromFile' with parameter `userdata_wl_content_type' against variable `TX:0' (Value: `text/plain' ) [file "/etc/nginx/modsec/comodo/10_HTTP_HTTP.conf"] [line "16"] [id "210710"] [rev "5"] [msg "COMODO WAF: Request content type is not allowed by policy. Please update file userdata_wl_content_type.||<SNIP>|F|2"] [data "REQUEST_METHOD=PUT"] [severity "2"] [ver ""] [maturity "0"] [accuracy "0"] [tag "CWAF"] [tag "HTTP"] [hostname "<SNIP>"] [uri "<SNIP>"] [unique_id "15953195826.827354"] [ref "v0,3o0,10o0,10v425,10"], client: <SNIP>, server: <SNIP>, request: "PUT <SNIP> HTTP/2.0", host: "<SNIP>"
```

The only reference to `NGX_LOG_INFO` is here: https://github.com/SpiderLabs/ModSecurity-nginx/blob/master/src/ngx_http_modsecurity_log.c#L33

I guess that `ngx_http_modsecurity_log` is used to log some of these actions.

Should I open a new issue?

PR #139, opened 2018-11-30
Module configuration refactoring

defanator:
The proposed changes are mostly aimed to improve overall code readability.

As a bonus, 2ae2a10 introduces a little feature of extending connector banner message with a total number of loaded rules with breakdown by rule source, e.g.:

```
2018/11/30 08:20:22 [notice] 9116#9116: ModSecurity-nginx v1.0.0 (rules loaded inline/local/remote: 0/1634/0)
```

### DIFF
--- a/src/ngx_http_modsecurity_body_filter.c
+++ b/src/ngx_http_modsecurity_body_filter.c
@@ -38,7 +38,7 @@ ngx_http_modsecurity_body_filter(ngx_http_request_t *r, ngx_chain_t *in)
     ngx_chain_t *chain = in;
     ngx_http_modsecurity_ctx_t *ctx = NULL;
 #if defined(MODSECURITY_SANITY_CHECKS) && (MODSECURITY_SANITY_CHECKS)
-    ngx_http_modsecurity_conf_t *loc_cf = NULL;
+    ngx_http_modsecurity_conf_t *mcf;
     ngx_list_part_t *part = &r->headers_out.headers.part;
     ngx_table_elt_t *data = part->elts;
     ngx_uint_t i = 0;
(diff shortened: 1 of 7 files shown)


zimmerle:
Hi @defanator,

The only thing that is crucial in terms of Rules merge, is the fact that somehow they have to respect the hierarchy in order to reduce the memory footprint. Two different virtual hosts can make usage of a rules set loaded in server configuration, instead of load two different copies of the rules into memory.

In the effort SpiderLabs/ModSecurity#1970 we are trying to allow the rules to be reloaded is run time, also, we are going to reduce the memory usage by replacing some textual tags with binary properties (good for memory and cpu) and shared_pointers instead of string repetition, pretty simple change that will produce a great result. But, yet, better to load the rules into memory only once, than have it loaded multiple times.

With that said, i think your patch is fine :) Indeed, it does improved the readability. The amount of loaded rules is amazing.

defanator:
@zimmerle rules merging hasn't changed with the proposed PR, I've just removed duplicate parts of code (i.e. merge_srv_conf and merge_loc_conf were actually doing the same stuff).

In regards to the dynamic modsecurity configuration management, I definitely like the idea. Currently one has to reload nginx by sending SIGHUP to master process after modsecurity configuration has changed; despite such approach can work without breaking live traffic (thanks to nginx' ability to gracefully pass incoming load to new set of workers), avoiding workers restart is a way better thing. It may bring a number of challenges though.

defanator:
@zimmerle in regards to memory footprint - yes, I'm pretty sure (lib)modsecurity (and/or connector modules) should be smart enough to detect whether the exactly same set of rules is being loaded for multiple entities (servers / locations / virtual hosts / you name it), and avoid any sort of duplicates.

I can imagine how to achieve this with loading rules from file (e.g. you can store and check SHA sums / modification times of a conf and its includes). Using remote sources may require additional checks like passing "If-Modified" and checking response codes, etc. Anyway, there is a room for improvements.

zimmerle:
Merged! ;) thanks!

Issue #142, opened 2018-12-10
SecRequestBodyLimit does not get properly merged from upper contexts


defanator:
Reproducible with the following test:

```
#!/usr/bin/perl

# (C) Andrei Belov

# Tests for ModSecurity-nginx connector (request body operations).

###############################################################################

use warnings;
use strict;

use Test::More;
use Socket qw/ CRLF /;

BEGIN { use FindBin; chdir($FindBin::Bin); }

use lib 'lib';
use Test::Nginx;

###############################################################################

select STDERR; $| = 1;
select STDOUT; $| = 1;

my $t = Test::Nginx->new()->has(qw/http auth_request/);

$t->write_file_expand('nginx.conf', <<'EOF');

%%TEST_GLOBALS%%

daemon off;

events {
}

http {
    %%TEST_GLOBALS_HTTP%%

    server {
        listen       127.0.0.1:8080;
        server_name  localhost;

        modsecurity on;
        modsecurity_rules '
            SecRuleEngine On
            SecRequestBodyAccess On
            SecRequestBodyLimit 128
            SecRequestBodyLimitAction Reject
            SecRule REQUEST_BODY "@rx BAD BODY" "id:31,phase:request,deny,log,status:403"
        ';

        client_header_buffer_size 1024;

        location /bodylimitreject {
#        modsecurity_rules '
#            SecRuleEngine On
#            SecRequestBodyAccess On
#            SecRequestBodyLimit 128
#            SecRequestBodyLimitAction Reject
#            SecRule REQUEST_BODY "@rx BAD BODY" "id:31,phase:request,deny,log,status:403"
#        ';
            proxy_pass http://127.0.0.1:8081;
        }
    }
}
EOF

$t->run_daemon(\&http_daemon);
$t->run()->waitforsocket('127.0.0.1:' . port(8081));

$t->plan(8);

###############################################################################

foreach my $method (('GET', 'POST', 'PUT', 'DELETE')) {
like(http_req_body($method, '/bodylimitreject', 'BODY' x 32), qr/TEST-OK-IF-YOU-SEE-THIS/, "$method request body limit reject, pass");
like(http_req_body($method, '/bodylimitreject', 'BODY' x 33), qr/403 Forbidden/, "$method request body limit reject, block");
}

###############################################################################

sub http_daemon {
	my $server = IO::Socket::INET->new(
		Proto => 'tcp',
		LocalHost => '127.0.0.1:' . port(8081),
[...]

PR #143, opened 2018-12-10
Tests: extend request body tests with case for issue #142

defanator:
While here, adjusted request body tests for flawless parallel execution.


defanator:
@zimmerle @victorhora I marked corresponding tests as TODO in order to avoid breaking of the whole suite - FYI. Those are being passed with https://github.com/SpiderLabs/ModSecurity/pull/1978 applied. We should remove TODO once the library is fixed.

zimmerle:
Merged! Thanks!

PR #173, opened 2020-01-09
Exit more gracefully if uri length is zero

martinhsv:


src/ngx_http_modsecurity_rewrite.c
```diff
@@ -150,6 +150,10 @@ ngx_http_modsecurity_rewrite_handler(ngx_http_request_t *r)
         if (n_uri == (char*)-1 || n_method == (char*)-1) {
```
defanator: I'm not sure whether the better way is to change this check in a way like `if (n_uri == NULL || n_method == NULL)`?

martinhsv: Assuming we want to leave the utility function ngx_str_to_char unchanged, it effectively has two different error return values:  NULL, and (char *)-1.  In that case it's useful to retain the existing line 150 check ... but then in addition also check for n_uri == NULL (the error use case that was occurring in https://github.com/SpiderLabs/ModSecurity/issues/2216).  We could of course also add a check for n_method==NULL, but I was focussed solely on improving the error handling for the case that we actually saw.

zimmerle:
This is merged! Thank you!

Issue #170, opened 2019-09-30
Requests blocked in Phase 1 (Request Headers) do not appear in the audit log


robinkb:
## Description

Requests that are blocked in Phase 1 (Request Headers) do not appear in the audit log. The rules are evaluated correctly, and block calls as expected. However, because they do not appear in the audit log, I am blind to a significant portion of possible disruptive actions in the ModSecurity CRS.

The issue is not present in the ModSecurity image based on Apache HTTPD. That is why I am reporting the issue here.

## Steps to reproduce

Start the ModSecurity CRS container based on NGINX.

```
$ docker run --rm -ti -p 8080:80 --name modsecurity-nginx \
      docker.io/owasp/modsecurity-crs:v3.2-modsec3-nginx
```

In another terminal, tail the audit log.

```
$ docker exec -ti modsecurity-nginx tail -f /var/log/modsec_audit.log
```

In yet another terminal, first confirm that the audit log works by triggering a warning. Then, confirm that the audit log does not work when the request is blocked by a rule in phase:1.

```
# Empty Accept header triggers a warning; returns status code 200.
$ curl http://localhost:8080/ -v -H "Accept;"

# Content-Length with invalid content is blocked; returns status code 400.
$ curl http://localhost:8080/ -v -H "Content-Length: not a digit" 
```

You can repeat the steps above with the Apache-based image, and see that the Apache-based image does not have this problem.

## Environment

I tried to track down what software versions were used in the Docker image, *which turned out to be quite difficult.*

The exact Docker image that I tested is [`docker.io/owasp/modsecurity-crs:v3.2-modsec3-nginx`](https://hub.docker.com/r/owasp/modsecurity-crs). Its Dockerfile is located [here](https://github.com/SpiderLabs/owasp-modsecurity-crs/blob/v3.2/dev/util/docker/Dockerfile-3.0-nginx), in the [`SpiderLabs/owasp-modsecurity-crs`](https://github.com/SpiderLabs/owasp-modsecurity-crs/tree/v3.2/dev) repository.

There, I can see that the image is based on [`docker.io/owasp/modsecurity:3.0-nginx`](https://hub.docker.com/r/owasp/modsecurity). The Dockerfile of *that* image appears to be located [here](https://github.com/CRS-support/modsecurity-docker/blob/v3/nginx-nginx/Dockerfile), in the [`CRS-support/modsecurity-docker`](https://github.com/CRS-support/modsecurity-docker) repository.

[...]

PR #175, opened 2020-01-27
Fix audit_log not generated for disruptive actions

victorhora:
PR intended to fix issues #170, SpiderLabs/ModSecurity/issues/2220 and SpiderLabs/ModSecurity/issues/2237 where the audit_log audit_log not generated for disruptive actions.

This changes commit https://github.com/SpiderLabs/ModSecurity-nginx/commit/baef88aac6c1eb3c132494f405a10166cbb50d6e that fixes the custom error pages issue (SpiderLabs/ModSecurity#2143) but it seemed to be returning too early not allowing the audit_logs to be generated on all phases.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -1,6 +1,8 @@
 v1.0.x - YYYY-MMM-DD (To be released)
 -------------------------------------
 
+ - Fix audit_log not generated for disruptive actions 
+   [Issue #170, #2220, #2237 - @victorhora]
  - Exit more gracefully if uri length is zero
    [@martinhsv]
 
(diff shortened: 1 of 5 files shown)


defanator:
@victorhora I have a strong feeling that we need tests here. Logging is clearly a PITA-point in the connector part as well, so having broader test coverage here will be a great benefit.

jeremyjpj0916:
Can confirm this PR does work, been testing it in my sandbox environments.

zimmerle:
@victorhora Merging this. It will be very good to have the tests as @defanator suggests.

Issue #187, opened 2020-03-26
client_body_in_file_only getting deleted


shmurf:
When setting 
`client_body_in_file_only on;`
nginx is supposed to cache the user request body to a file and NOT delete it, this is convenient for debugging.
However with modsec on I realized that the file gets deleted even when set to ON.

martinhsv:
Hello @shmurf ,

Thank you for raising that.

I can confirm your report ... and I think I see the problem.

zimmerle:
Closed as #194 got merged! I was expecting GitHub to do it automagically :P

PR #192, opened 2020-04-08
Fix client_body_in_file_only config setting not respected

martinhsv:


defanator:
Hi @martinhsv, any story behind this change?

defanator:
I mean, if #187 was the reason, then the `request_body_in_clean_file` flag should not be touched:
https://github.com/nginx/nginx/blob/master/src/http/ngx_http_core_module.c#L1316-L1317

I'm not sure about another couple, as I barely remember there were some issues with connector tests. Happy to be wrong, but it's worth to check.

martinhsv:
Hi @defanator ,

The line:
```
r->request_body_in_clean_file = 1;
```
cannot be left in place or the current issue will not be fixed.

The nginx code that you have pointed to sets request_body_in_clean_file to 1 only if the configuration file has 'client_body_in_file_only clean;'

If the config file has 'client_body_in_file_only on;' as in the poster's case, then the nginx code will set the variable request_body_in_clean_file=0 ... and then the ModSecurity connector code will effectively override that and always set it to 1.

Everything I've seen suggests that the three line-deletes in this pull request should proceed, since those lines being present amount to arbitrarily override nginx's own configuration setup.  (I did also note the code comments associated with those three lines ... the original coder obviously wasn't confident that they ought to be there.)

Nevertheless, if there is a particular configuration+use case that you're worried won't work correctly with this pull request, feel free to let me know and I'll be happy to try it out.

defanator:
@martinhsv well, the proposed patch is breaking the following tests:

```
Test Summary Report
-------------------
modsecurity-h2.t             (Wstat: 0 Tests: 22 Failed: 0)
  TODO passed:   21
modsecurity-proxy-h2.t       (Wstat: 0 Tests: 25 Failed: 0)
  TODO passed:   24
modsecurity-request-body-h2.t (Wstat: 3072 Tests: 38 Failed: 12)
  Failed tests:  2, 6, 8, 10, 14, 16, 18, 22, 24, 26, 30
                32
  Non-zero exit status: 12
Files=13, Tests=230, 34 wallclock secs ( 0.07 usr  0.02 sys +  1.06 cusr  0.24 csys =  1.39 CPU)
Result: FAIL
```

I did a quick check on past issues both in library and nginx connector to bring more context, but haven't found anything significant (I'll also check our internal KB/support cases when time permits).

defanator:
It seems like from all the following

```
         r->request_body_in_single_buf = 1;
         r->request_body_in_persistent_file = 1;
         r->request_body_in_clean_file = 1;
```

only `r->request_body_in_persistent_file = 1` is really essential (from tests point of view).

I think that removing another couple would resolve #187 and keep tests happy.

Thoughts?

martinhsv:
Hi @defanator ,

Thanks for that.  I'm a bit surprised by the result, but good to know.

I'll alter/replace this pull request with one where the only change is to delete the single line 'r->request_body_in_clean_file = 1;'

That's probably the safest thing to do for now since 1) That's the only change that I've identified as absolutely necessary to resolve the current issue and 2) You've already identified some side-effects to a broader change.

martinhsv:
Closing this in favour of pull request #194

Issue #182, opened 2020-02-20
ModSecurity False Positive with respect to logging in ModSec audit + NGX error logs due to NGINX behavior with error_page directive


jeremyjpj0916:
Decided to pull out my comment to a separate issue since its unrelated to the other issue I raised originally:

I was trying to limit the relevant to 400's and 403's for capturing blocked transactions by ModSecurity:

```SecAuditLogRelevantStatus "^(?:4((00)|(03)))"```

Yet I see the audit log in RelevantOnly mode still also capturing 504 status messages as well for some reason.

SO this one is an interesting case. I am 100% sure on why 504's were making it into the audit logs even through its regex was set to 400's and 403's because I saw odd behavior in other apps too that hook into NGX with logged values not matching what the client really sent.

Firstly take a gander here(the app we run on nginx + openresty), leveraging the error_page directive:
https://github.com/Kong/kong/blob/master/kong/templates/nginx_kong.lua#L83

Then take a gander here at how Kong or OpenResty can be patched for such situations:
Kong/kong#5073
openresty/openresty#396

So what happens is when NGINX does an internal redirect via error_page, and the CTX of the client HTTP METHOD gets reset to GET always, so the NGINX ModSec audit/error logs are thinking REAL 504 timeout transactions are now a GET with a HTTP body(client sends req body) in the audit log logic(The Method the client actually used in the originating request was a POST), so it saves an blocked error log w 403 in the audit log file section H, for the transaction BUT the real transaction was a 504 and it was presented to the client as a 504, just internal logging and record keeping Modsecurity-nginx is not handling these scenarios where error_page directive gets used by folks leveraging the NGINX web server. 

The fix could easily be if yall added a custom mod security http method variable to the tx context in the pre-access phase taking the original ```r->method``` to track the Method before the error_page directive internal redirect and read that custom mod sec http method variable after in the audit/error log phase logic after the redirect has occurred as the source of truth rather than the current ngx contexts ```r->method```(which would be reset to always be GET at that time). Will save on the noisy invalid logs.

Ex of current behavior(Upstream times out in reverse proxy call, we hit error_page directive, http method gets reset to GET internally, triggering ModSec to log this tx as a block in error + audit log incorrectly when it was really just a harmless timeout):

> 
[...]

PR #204, opened 2020-05-22
Fix error_page internal redirect false positive logging

jeremyjpj0916:
Fixes: #182 

Problem here was ```error_page``` redirects don't maintain original client request data. Hence ModSecurity has false information when evaluating the clients HTTP request a second time around during the request(HTTP Method Header is always reset to GET, URI is also not the original from client request). It is also inefficient to evaluate the request data a second time around after the internal redirect, hence cutting out those phases the second time around after the redirect makes sense and is a performance improvement.

To figure it out I was reviewing prior commits, specifically:

https://github.com/SpiderLabs/ModSecurity-nginx/commit/ce1d438bc665c28853eb5b429aad515d42027919

Which regarding this commit did indeed ensure the log phase still runs during ```error_page``` internal redirect, but just had not considered the implications of re-processing the NGX request phases(phases prior to the ```proxy_pass``` directive for instance) again.

Currently running these changes in our dev and stage environments personally. Will be pushing to our prod in the next week or so, will run as a fork until this is merged.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -1,9 +1,10 @@
 v1.0.x - YYYY-MMM-DD (To be released)
 -------------------------------------
-
+ - Fix request logging false positives due to error_page internal redirects
+   [Issue #182 - @jeremyjpj0916]
  - Fix config setting not respected: client_body_in_file_only on
    [Issue #187 - @martinhsv]
- - Fix audit_log not generated for disruptive actions 
+ - Fix audit_log not generated for disruptive actions
    [Issue #170, #2220, #2237 - @victorhora]
  - Exit more gracefully if uri length is zero
    [@martinhsv]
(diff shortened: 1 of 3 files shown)


jeremyjpj0916:
@zimmerle @martinhsv @defanator @victorhora  Any thoughts here? 

Should be good to ```#ShipIt````

jeremyjpj0916:
still pending? oof.

zimmerle:
@defanator Can you have a look at this?

jeremyjpj0916:
Github bot being bad again 💀.

jeremyjpj0916:
Can say I have ran this in my fork of code for a long time 10/10 does indeed work 😆 .

defanator:
@jeremyjpj0916 copying here as well - it is believed that https://github.com/SpiderLabs/ModSecurity-nginx/commit/4f26b48998db5119ca818b0909b6b14e08ebb544 should resolve at least some of observed misbehaviours reported in #182.

If it doesn't, I would appreciate a test case demonstrating the issue with minimal configuration involved.

jeremyjpj0916:
What was wrong with the simple original change from my PR to ignore processings on redirects since ModSecurity will have the wrong context post an error page redirect?

PR #225, opened 2020-10-02
Support http protocol versions besides 0.9, 1.0, 1.1, 2.0

martinhsv:
This pull request was prompted by https://github.com/SpiderLabs/ModSecurity/issues/2380.

Until this change, any HTTP protocol version other than 0.9, 1.0, 1.1, and 2.0 would actually result in the ModSecurity-nginx connector passing a value of "1.0" to ModSecurity's msc_process_uri function.

With this change, any characters following the five-character 'HTTP/' protocol prefix will get passed to ModSecurity, as long as nginx has successfully populated ngx_http_request_t->http_protocol.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -1,6 +1,8 @@
 v1.0.x - YYYY-MMM-DD (To be released)
 -------------------------------------
 
+ - Support http protocol versions besides 0.9, 1.0, 1.1, 2.0
+   [Issue #224 - @HQuest, @martinhsv]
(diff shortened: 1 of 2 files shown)


flo-mic:
Are there any updates on this? What is blocking this pull request?

Would like to see this merged to get http3 working. There is already a dedicated nginx branch for quic support but it can not be used as long as http3 is not supported by the ModScurity-nginx module.

See https://hg.nginx.org/nginx-quic

PR #243, opened 2021-03-22
Adjust nginx log level according to modsecurity severity level

qix67:
Without this patch, whatever severity level in modsecurity message, it
appears as 'error' in nginx error.log file.

This patch modifies this behavior. When a severity level is present, it
is used as nginx log level else 'error' nginx log level is used.


martinhsv:
Hello @qix67 ,

There are a few open issues with a somewhat similar goal (by which I mean providing additional control over the log level of lines written to error.log).

In general I'm at least somewhat hesitant to equate ModSecurity's rule 'severity' with the log level in a web server's error.log file. (See https://github.com/SpiderLabs/ModSecurity/pull/2748#issuecomment-1231058099 for some reasoning.)

Moreover, *if* we did decide to do something along those lines, I think it would be preferable to implement it through a specific variable, rather than parsing log lines. Parsing log line content has a chance of not producing the intended results. I haven't tried the proposed code here, but, for example, what if the 11-character string '[severity "' were to appear somewhere else in the rule output? Perhaps the admin, for obscure reasons included text like that in the rule's "msg" action?

PR #271, opened 2022-03-04
Stop segfaults caused by NULL request_body

brandonpayton:
ModSecurity-nginx assumes `ngx_http_request_t.request_body` is never NULL and encounters a segfault when the `request_body` is in fact NULL.

We have seen this happen when ModSecurity-nginx is used in conjunction with [lua-nginx-module](https://github.com/openresty/lua-nginx-module). When a subrequest is made using [this lua API](https://openresty-reference.readthedocs.io/en/latest/Lua_Nginx_API/#ngxlocationcapture), it can result in the `request_body` being set to NULL [here](https://github.com/openresty/lua-nginx-module/blob/ccc748eafbab96d0795dc92dca9374227e4b46f5/src/ngx_http_lua_subrequest.c#L646-L651).

I considered whether this was more of a bug with lua-nginx-module, but nginx's codebase appears to recognize a NULL `request_body` is possible (some examples: [a](https://github.com/nginx/nginx/blob/828fb94e1dbe1c433edd39147ba085c4622c99ed/src/http/ngx_http_variables.c#L2059), [b](https://github.com/nginx/nginx/blob/828fb94e1dbe1c433edd39147ba085c4622c99ed/src/http/modules/ngx_http_auth_request_module.c#L202), [c](https://github.com/nginx/nginx/blob/828fb94e1dbe1c433edd39147ba085c4622c99ed/src/http/modules/ngx_http_dav_module.c#L216)). Also, it seems reasonable for ModSecurity-nginx to be a little more defensive in this case.

This is a patch to fix that issue by wrapping the code that process the request_body in an if-NULL check. With this patch, msc_append_request_body() will not be called when the `request_body` is NULL, and this seems like it will be OK because ModSecurity's Transaction's [m_requestBody is a std::ostringstream](https://github.com/SpiderLabs/ModSecurity/blob/5519f6cfae45a47d8dc3ac1b084319611a9b386b/headers/modsecurity/transaction.h#L549) that will simply not have any data.

### DIFF
--- a/src/ngx_http_modsecurity_pre_access.c
+++ b/src/ngx_http_modsecurity_pre_access.c
@@ -138,64 +138,67 @@ ngx_http_modsecurity_pre_access_handler(ngx_http_request_t *r)
     if (ctx->waiting_more_body == 0)
     {
         int ret = 0;
-        int already_inspected = 0;
-
-        dd("request body is ready to be processed");
-
-        r->write_event_handler = ngx_http_core_run_phases;
-
-        ngx_chain_t *chain = r->request_body->bufs;
-
-        /**
-         * TODO: Speed up the analysis by sending chunk while they arrive.
-         *
-         * Notice that we are waiting for the full request body to
-         * start to process it, it may not be necessary. We may send
-         * the chunks to ModSecurity while nginx keep calling this
-         * function.
-         */
-
-        if (r->request_body->temp_file != NULL) {
-            ngx_str_t file_path = r->request_body->temp_file->file.name;
-            const char *file_name = ngx_str_to_char(file_path, r->pool);
-            if (file_name == (char*)-1) {
-                return NGX_HTTP_INTERNAL_SERVER_ERROR;
-            }
-            /*
-             * Request body was saved to a file, probably we don't have a
-             * copy of it in memory.
-             */
-            dd("request body inspection: file -- %s", file_name);
-
-            msc_request_body_from_file(ctx->modsec_transaction, file_name);
-
-            already_inspected = 1;
-        } else {
-            dd("inspection request body in memory.");
-        }
-
-        while (chain && !already_inspected)
+        if (r->request_body != NULL)
         {
-            u_char *data = chain->buf->pos;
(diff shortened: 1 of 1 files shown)


brandonpayton:
In the case of this patch, viewing with ignoring whitespace changes emphasizes the simplicity of this change:
https://github.com/SpiderLabs/ModSecurity-nginx/pull/271/files?w=1

On another note, I just realized that [nginx checks r->request_body->bufs for NULL](https://github.com/nginx/nginx/blob/828fb94e1dbe1c433edd39147ba085c4622c99ed/src/http/ngx_http_variables.c#L2060) when checking `request_body`, so I plan to add that the same sort of check to this PR when back at my test machine.

brandonpayton:
Unfortunately, reproducing the segfault appears to require building nginx with a module that sets request_body to NULL and triggering it, but perhaps it would be sufficient to observe that making sure request_body is not NULL does not break anything.

On my test machine, I've built nginx with ModSecurity-nginx and lua-nginx-module and added a location to the nginx config that triggers setting request_body to NULL. With that setup, I am able to reproduce the segfault and confirm that, with this patch, the segfault no longer occurs.

brandonpayton:
> On another note, I just realized that [nginx checks r->request_body->bufs for NULL](https://github.com/nginx/nginx/blob/828fb94e1dbe1c433edd39147ba085c4622c99ed/src/http/ngx_http_variables.c#L2060) when checking request_body, so I plan to add that the same sort of check to this PR when back at my test machine.

Actually, this does not appear to be strictly needed because `r->request_body->bufs` is assigned to `chain` [here](https://github.com/SpiderLabs/ModSecurity-nginx/blob/2497e6ac654d0b117b9534aa735b757c6b11c84f/src/ngx_http_modsecurity_pre_access.c#L147), and `chain` is not used without checking it first [here](https://github.com/SpiderLabs/ModSecurity-nginx/blob/2497e6ac654d0b117b9534aa735b757c6b11c84f/src/ngx_http_modsecurity_pre_access.c#L177).

PR #273, opened 2022-03-11
recovery context after internal redirect

liudongmiao:
In `error_page`, nginx would reset context. Then `modsecurity-nginx` cannot recovery from previous context.

It will act like this:
- phase 0 (connection)
- phase 1 (request headers)
- phase 2 (request body)

Then, like a new request: (request method set to `GET`)
- phase 0 (connection)
- phase 1 (request headers)
- phase 2 (request body)
- ... other ... phases

In previous, `r->error_page` was introduced to solve the `log` problem. However, it doesn't solve the root cause.

This PR should close #270, #255 and other `error_page` (internal redirect) issues.

### DIFF
--- a/src/ngx_http_modsecurity_body_filter.c
+++ b/src/ngx_http_modsecurity_body_filter.c
@@ -48,7 +48,7 @@ ngx_http_modsecurity_body_filter(ngx_http_request_t *r, ngx_chain_t *in)
         return ngx_http_next_body_filter(r, in);
     }
 
-    ctx = ngx_http_get_module_ctx(r, ngx_http_modsecurity_module);
+    ctx = ngx_http_modsecurity_get_module_ctx(r);
 
     dd("body filter, recovering ctx: %p", ctx);
 
--- a/src/ngx_http_modsecurity_common.h
+++ b/src/ngx_http_modsecurity_common.h
@@ -99,6 +99,7 @@ typedef struct {
     unsigned processed:1;
     unsigned logged:1;
     unsigned intervention_triggered:1;
+    unsigned request_body_processed:1;
 } ngx_http_modsecurity_ctx_t;
 
 
@@ -139,6 +140,7 @@ extern ngx_module_t ngx_http_modsecurity_module;
 /* ngx_http_modsecurity_module.c */
 int ngx_http_modsecurity_process_intervention (Transaction *transaction, ngx_http_request_t *r, ngx_int_t early_log);
 ngx_http_modsecurity_ctx_t *ngx_http_modsecurity_create_ctx(ngx_http_request_t *r);
+ngx_http_modsecurity_ctx_t *ngx_http_modsecurity_get_module_ctx(ngx_http_request_t *r);
 char *ngx_str_to_char(ngx_str_t a, ngx_pool_t *p);
 ngx_pool_t *ngx_http_modsecurity_pcre_malloc_init(ngx_pool_t *pool);
 void ngx_http_modsecurity_pcre_malloc_done(ngx_pool_t *old_pool);
--- a/src/ngx_http_modsecurity_header_filter.c
+++ b/src/ngx_http_modsecurity_header_filter.c
@@ -107,7 +107,7 @@ ngx_http_modsecurity_store_ctx_header(ngx_http_request_t *r, ngx_str_t *name, ng
     ngx_http_modsecurity_conf_t    *mcf;
     ngx_http_modsecurity_header_t  *hdr;
 
-    ctx = ngx_http_get_module_ctx(r, ngx_http_modsecurity_module);
+    ctx = ngx_http_modsecurity_get_module_ctx(r);
     if (ctx == NULL || ctx->sanity_headers_out == NULL) {
         return NGX_ERROR;
     }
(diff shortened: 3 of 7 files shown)


liudongmiao:
@martinhsv could you help to review for this?

liudongmiao:
@zimmerle @AirisX @defanator 
Could you help to check for this?

#241 doesn't resolve the problem too.

liudongmiao:
@AirisX I think https://github.com/SpiderLabs/ModSecurity-nginx/blob/master/tests/modsecurity-config-custom-error-page.t should be ok. However, the test case doesn't check the created transaction in log. And if the request method is non-HEAD, it would be changed to GET.

In my test case, without this patch, there would be two transaction. If you don't use custom erorr_page, you can ignore this.

For erorr_page, nginx would call `ngx_http_send_error_page` in `src/http/ngx_http_special_response.c`, then nginx would call `ngx_memzero` in `ngx_http_internal_redirect` and `ngx_http_named_location` (for named location).

You can debug with IDE like Clion, or read http://nginx.org/en/docs/dev/development_guide.html#http_request_redirection

- `ngx_http_internal_redirect` would clear context, then start from phase `NGX_HTTP_SERVER_REWRITE_PHASE`
- `ngx_http_named_location` would clear context, then start from phase `NGX_HTTP_REWRITE_PHASE`

All nginx phases:
- NGX_HTTP_POST_READ_PHASE = 0,
- NGX_HTTP_SERVER_REWRITE_PHASE,
- NGX_HTTP_FIND_CONFIG_PHASE,
- NGX_HTTP_REWRITE_PHASE, `<-- modsecurity hook, create new transaction if cannot get from context`
- NGX_HTTP_POST_REWRITE_PHASE,
- NGX_HTTP_PREACCESS_PHASE, `<-- modsecurity hook for request body`
- NGX_HTTP_ACCESS_PHASE,
- NGX_HTTP_POST_ACCESS_PHASE,
- NGX_HTTP_PRECONTENT_PHASE,
- NGX_HTTP_CONTENT_PHASE,
- NGX_HTTP_LOG_PHASE `<-- modsecurity hook for log`

mlevogiannis:
We also agree that this pull request is in the right direction, however it does not directly close pull request #255, contrary to what is stated in the first comment. Specifically, the log handler will only run if ModSecurity is also enabled in the internal redirect's location conf, but not if it is not (with the exception of phase 1 rules where the log handler runs because of the "early logging" workaround). This is demonstrated by the updated test case included in pull request #255 (the [current](https://github.com/SpiderLabs/ModSecurity-nginx/blob/v1.0.2/tests/modsecurity-config-custom-error-page.t) test case does not cover the aforementioned problematic cases).

The problem is that the connector decides whether ModSecurity is enabled or not based on the configuration directive in the request's location conf. At the time the log handler runs, the original request has already been replaced by a new one with the internal redirect's location conf. As a result, if ModSecurity is not enabled in the internal redirect's location conf, the check at [line 47](https://github.com/SpiderLabs/ModSecurity-nginx/blob/v1.0.2/src/ngx_http_modsecurity_log.c#L47) will fail and the original transaction will not be logged.

What is required is to make this decision based on the original request, in which case the recovered context should be used instead of the location conf. There are two possible options here:

- Create a new context in the rewrite handler regardless of the configuration directive in the location conf and use an extra flag in the context to mark whether ModSecurity is enabled or not. This flag will then be checked by the pre access handler, the header filter, the body filter and the log handler after the original context has been recovered.
- Extend the current behavior of the header and body filters, which assume that a null context means that ModSecurity is disabled, to the pre access and log handlers. This way the log handler will run regardless of ModSecurity being enabled or not in the internal redirect's location conf.

As a PoC, the second option was implemented in [this](https://github.com/grnet/ModSecurity-nginx/tree/bugfix/recover-context-after-internal-redirect) branch. These changes can either be included in this pull request, or we can open a new one.

SWGAKamui:
Is there any news please?
It seems that those commits helps my need.
Thanks to the author for your contribution. :)

liudongmiao:
@mlevogiannis I have checked your commit, how to merge it into this PR?

lmq1999:
This commit should be merge, I having same prob

jessp01:
Hi all,

Just wondering where things stand with this?

Thanks,

airween:
To all: sorry for the delay, the new team takes care with sources since end of January, 2024. We weren't able to review all PR's, but now I try to take a look here too.

Thanks and sorry again.

mlevogiannis:
I have rebased onto the latest master and updated [the branch in our repository](https://github.com/grnet/ModSecurity-nginx/tree/bugfix/recover-context-after-internal-redirect) so that the newly introduced CI workflow tests pass. As a recap, our branch:
* Includes the original commit of this PR, which properly recovers the module's context in case of an internal redirect.
* Uses the recovered context to decide whether ModSecurity is enabled or not, which properly fixes #255.
* Adds test cases that cover the issues described in #255.
* Removes the partial "early logging" workaround, which is no longer required.

@liudongmiao Can you please update this PR's branch with the changes in our branch? You may either cherry-pick the 3 extra commits or "git reset" your branch with our branch and force push it.

liudongmiao:
@mlevogiannis The original patch should be very easy to understand, as it will get the context from clean handler if the context is reset. And there is reference from realip module.

jeremyjpj0916:
Definitely need to see the error_page issues with ModSecurity nginx resolved after 2 years of it being in limbo and all of us running patch files to achieve functionality that should be baked into the main repo for all to benefit.

lmq1999:
should this problem need to solve to main branches ?

airween:
Please read [this](https://github.com/owasp-modsecurity/ModSecurity-nginx/issues/314#issuecomment-2419310766) comment. I need some help to reproduce the original issue.

liudongmiao:
merged in #326

PR #275, opened 2022-03-21
use msc_init to check modsecurity

liudongmiao:
printf is not part of modsecurity, and may not work with -Werror

### DIFF
--- a/config
+++ b/config
@@ -12,7 +12,7 @@ ngx_feature_name=
 ngx_feature_run=no
 ngx_feature_incs="#include <modsecurity/modsecurity.h>"
 ngx_feature_libs="-lmodsecurity"
-ngx_feature_test='printf("hello");'
+ngx_feature_test='msc_init();'
 ngx_modsecurity_opt_I=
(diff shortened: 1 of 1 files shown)


mikelolasagasti:
This patch is required by downstream Fedora. It is required even if `MODSECURITY_LIB` and `MODSECURITY_INC` are set.

airween:
@liudongmiao thanks, this is very useful. (I had [solved](https://salsa.debian.org/modsecurity-packaging-team/libnginx-mod-http-modsecurity/-/blob/master/debian/patches/stdioh.patch?ref_type=heads) that with a bit similar way in Debian - but this one is more elegant.)

Could you pick up the current status and rebase your tree? It would be nice to run every PR's against our new workflow.

Thanks!

liudongmiao:
@airween Thanks. 

PR #275 was made two years ago. In last 2 years, modsecurity and modsecurity-nginx have changed.

liudongmiao:
close as fixed in #352

PR #289, opened 2022-08-30
Put the ModSecurity interception logic onto separated threads

### LINKED ISSUE #227: Synchronous process of the request destroys performance
As mentioned in [nginx dev guide](http://nginx.org/en/docs/dev/development_guide.html#libraries) using blocking libraries destroys performance. For example after enabling Modsecurity module, the performance of nginx decreases about 70%.

wfjsw:
This PR is created to describe my patch to fix #227 and it is by no means a complete patch ready for merge. 

The patch contains several unrelated changes, namely:

- Logger change, from `dd` to `ngx_log_error` to accomodate my own debugging need 
- Removal of `ngx_http_modsecurity_pcre_malloc_init` and `ngx_http_modsecurity_pcre_malloc_done`. They are not used in my configuration where PCRE2 is used, and it looks suspicious for SEGVs so I commented them out as a precaution.
- Why use `NGX_OK` in logging handler? Changed to `NGX_DECLINED`.

Not yet implemented:

- I'm not sure whether logging phase would take advantage of this. I did not change that yet.
- Missing a `NGX_THREADS` guard for Nginx setup without threading support.

Currently it passes all test suites and performs well in production.

Benchmarking is welcomed.

### DIFF
--- a/src/ngx_http_modsecurity_body_filter.c
+++ b/src/ngx_http_modsecurity_body_filter.c
@@ -50,7 +50,7 @@ ngx_http_modsecurity_body_filter(ngx_http_request_t *r, ngx_chain_t *in)
 
     ctx = ngx_http_get_module_ctx(r, ngx_http_modsecurity_module);
 
-    dd("body filter, recovering ctx: %p", ctx);
+    ngx_log_error(NGX_LOG_DEBUG, r->connection->log, 0, "body filter, recovering ctx: %p", ctx);
 
     if (ctx == NULL) {
         return ngx_http_next_body_filter(r, in);
@@ -155,11 +155,11 @@ ngx_http_modsecurity_body_filter(ngx_http_request_t *r, ngx_chain_t *in)
         is_request_processed = chain->buf->last_buf;
 
         if (is_request_processed) {
-            ngx_pool_t *old_pool;
+            // ngx_pool_t *old_pool;
 
-            old_pool = ngx_http_modsecurity_pcre_malloc_init(r->pool);
+            // old_pool = ngx_http_modsecurity_pcre_malloc_init(r->pool);
             msc_process_response_body(ctx->modsec_transaction);
-            ngx_http_modsecurity_pcre_malloc_done(old_pool);
+            // ngx_http_modsecurity_pcre_malloc_done(old_pool);
 
 /* XXX: I don't get how body from modsec being transferred to nginx's buffer.  If so - after adjusting of nginx's
    XXX: body we can proceed to adjust body size (content-length).  see xslt_body_filter() for example */
@@ -176,7 +176,7 @@ ngx_http_modsecurity_body_filter(ngx_http_request_t *r, ngx_chain_t *in)
     }
     if (!is_request_processed)
     {
-        dd("buffer was not fully loaded! ctx: %p", ctx);
+        ngx_log_error(NGX_LOG_DEBUG, r->connection->log, 0, "buffer was not fully loaded! ctx: %p", ctx);
     }
 
 /* XXX: xflt_filter() -- return NGX_OK here */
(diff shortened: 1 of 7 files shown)


airween:
Hi @wfjsw, there is a new CI workflow test in this repository. Could you pick up the modifications to enable run those tests? Thanks!

jeremyjpj0916:
I wonder if multi threading is at least the partial answer to fix the long known problem of ModSecurity+nginx being a performance killer for throughput.

wfjsw:
I no longer have a ModSecurity install on my machine so I'm unable to investigate further :(

SonNgo2211:
Hi @airween,
I have read this blog post: https://blog.litespeedtech.com/2021/02/22/asynchronous-modsecurity-engine/
It mentions the asynchronous ModSecurity engine and how request blocking is handled.
Has this topic been included in the roadmap or is there any related plan for further enhancements?
Thanks!

airween:
Hi @SonNgo2211,

> Hi @airween, I have read this blog post: https://blog.litespeedtech.com/2021/02/22/asynchronous-modsecurity-engine/ It mentions the asynchronous ModSecurity engine and how request blocking is handled. Has this topic been included in the roadmap or is there any related plan for further enhancements? Thanks!

Thanks for this link. I'm not a LiteSpeed expert (I haven used that at all yet), but I think the key is this:
_"As of LiteSpeed Web Server v6.0, **we have introduced a separate worker thread pool for ModSecurity processing**."_

As I understand that the solution is based on a web server (LiteSpeed) function. So there is no similar function in Nginx, I'm afraid there are no plans to use this mechanism in Nginx.

SonNgo2211:
I noticed that Nginx supports asynchronous thread pools for offloading blocking main working cycle (see: https://www.nginx.com/blog/thread-pools-boost-performance-9x/).

As far as I know, ModSecurity/ModSecurity-Nginx connector run synchronously within the Nginx worker’s event loop. This means heavy or complex rules can still block the worker, and the Nginx thread pools cannot currently be used to offload WAF evaluation away from the main event loop.

This PR discussing separated threads for ModSecurity (similar solution of LiteSpeed). Is there any technical constraint or design decision that prevents such an approach in connector Modsecurity-Nginx?  I’d love to learn more. Thanks @airween!

SonNgo2211:
> I noticed that Nginx supports asynchronous thread pools for offloading blocking main working cycle (see: https://www.nginx.com/blog/thread-pools-boost-performance-9x/).

Sorry for my initial confusion!
I read the blog more thoroughly and did some more digging. It seems that Nginx thread pools are only used to offload blocking disk I/O operations (such as file reads/writes and sendfile operations).

I would still be interested to learn if there are any future plans/discussions about supporting asynchronous/multi-threaded WAF assessments in the ModSecurity-nginx connector. I will continue to monitor the issue https://github.com/owasp-modsecurity/ModSecurity/issues/3215

airween:
> I would still be interested to learn if there are any future plans/discussions about supporting asynchronous/multi-threaded WAF assessments in the ModSecurity-nginx connector.

Thanks again for bringing up this topic! Actually there is no any plan/discussion about async/mt threaded developing, but this does not mean we don't want to discuss it.

Currently I'm working on the libmodsecurity3's API changes (it's necessary to change the current API), then I have to align the connector for it. Probably the connector version will be 1.1.0 then.

After that, we can start to think about this topic.

SonNgo2211:
Thanks a lot @airween! 
It's great to see that the topic of asynchronous/multithreaded WAF is at least still open for future discussion. I hope to have a lot of fun with the upcoming API release in libmodsecurity3! As a Modsecurity user I'm happy to contribute use cases and issues from real implementations to help it grow.

PR #291, opened 2022-11-16
Reuse already initialized rule set for each location

k1k:
Only merge rule-sets when there are extra rules in the location, otherwise just re-use the rule-set of the parent


airween:
Hi @k1k, we've added a CI workflow to this repository. Could you pick up the modifications and send your patch again? It would be fine to run those tests - thanks!

HanadaLee:
The lazy/shared ruleset approach is reasonable, but it changes the behavior of modsecurity on without any loaded rules. Both parent and child rulesets can remain NULL, ngx_http_modsecurity_create_ctx() then returns NULL, and the rewrite handler turns that into HTTP 500. Previously every location had a valid empty RulesSet.

Please preserve the empty-ruleset behavior for enabled configurations. Also, the DDEBUG dumps currently run before the new NULL checks, while msc_rules_dump() dereferences its argument unconditionally, causing debug builds to crash during configuration merging.

tomsommer:
Confirming the issue raised in the review: with this change both `p->rules_set` and `c->rules_set` stay NULL for an enabled configuration that has no `modsecurity_rules*` directive anywhere (for example `modsecurity on;` at `http` level with rules meant to be added later, or a server that only inherits), `ngx_http_modsecurity_create_ctx()` then returns NULL and every request gets a 500. Before the change every location had a valid empty `RulesSet`.

I have a variant of this idea that keeps that behaviour: rule sets are created lazily in the three `modsecurity_rules*` handlers (one pool cleanup per created set), `merge_conf()` shares the parent's pointer when the child has no rules of its own and merges into the child's own set otherwise, and enabled configurations with no rules anywhere share a single empty set owned by the main conf. It comes with a Test::Nginx file covering inheritance, own+parent, nested locations, `if`/`limit_except` blocks, `modsecurity off` and the no-rules case, passes the existing suite on libmodsecurity 3.0.9 and 3.0.14, and survives repeated `-s reload` under valgrind.

Branch: https://github.com/tomsommer/ModSecurity-nginx/tree/perf/share-rules-set (single commit). Happy for you to cherry-pick it into this PR, or I can open it as a separate PR if the maintainers prefer. Prepared during a performance review of the connector done with Claude Fable 5.1.

PR #317, opened 2024-04-11
feat: added Github CI workflow

airween:
Added `.github/workflow/ci.yml` and necessary files with a complex test workflow for each PR.

A brief summary about how does it work:
* set up dependencies (required packages)
* install latest ModSecurity3 release
* get ModSecurity-nginx source
* get latest Nginx soure
* build Nginx with ModSecurity module
* start Nginx with ModSecurity
* there are two virtual hosts
* there is only one rule
* run a valid request (expected code is 200) and an attack (expected code is 403) against both vhosts


.github/mime.types
theseion, line 1: I would move the files needed for ngnix into a separate directory called `nginx` (`.github/nginx`). That's just personal preference though.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,65 @@
...
+        platform: [x32, x64]
+        compiler: [gcc, clang]
+    steps:
+      - name: Setup Dependencies
+        run: |
+          sudo apt-get update -y -qq
+          sudo apt-get install -y gcc g++ make autoconf automake make libyajl-dev libxml2-dev libmaxminddb-dev libpcre2-dev libpcre2-8-0 libpcre2-16-0 libpcre2-32-0 libcurl4-gnutls-dev jq wget
+      - name: Install ModSecurity library
+        run: |
+          modsec=`curl -s https://api.github.com/repos/owasp-modsecurity/ModSecurity/releases/latest | jq .assets[0].browser_download_url | tr -d "\""`
```
theseion, line 22: Use the GitHub CLI. It's available on every runner by default. e.g.:

```yaml
     - name: "Install dependencies"
        env:
          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
        run: |
          gh release download -R coreruleset/go-ftw "v${{ env.GO_FTW_VERSION }}" \
            -p "ftw_${{ env.GO_FTW_VERSION }}_linux_amd64.tar.gz" -O - | tar -xzvf - ftw
```

.github/workflows/ci.yml
```diff
@@ -0,0 +1,65 @@
...
+          modsecver=`curl -s https://api.github.com/repos/owasp-modsecurity/ModSecurity/releases/latest | jq .name | tr -d "\""`
+          wget $modsec -O modsecurity.tar.gz
+          tar -xzf modsecurity.tar.gz
+          cd modsecurity-$modsecver
+          ./configure --disable-lmdb --prefix=/usr
+          make -j `nproc`
```
theseion, line 28:
```suggestion
          make -j $(nproc)
```

.github/workflows/ci.yml
```diff
@@ -0,0 +1,65 @@
...
+          sudo make install
+          cd
+      - uses: actions/checkout@v2
```
theseion, line 31: Use versions tagged to hashes. You can later set up Dependabot or Renovate to automatically update these dependencies in your workflows periodically.

```suggestion
      - uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
```

.github/workflows/ci.yml
```diff
@@ -0,0 +1,65 @@
...
+        with:
+          path: ModSecurity-nginx
+      - name: Get Nginx source
+        uses: actions/checkout@v3
```
theseion, line 35:
```suggestion
        uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
```

.github/workflows/ci.yml
```diff
@@ -0,0 +1,65 @@
...
+          wget $modsec -O modsecurity.tar.gz
+          tar -xzf modsecurity.tar.gz
+          cd modsecurity-$modsecver
+          ./configure --disable-lmdb --prefix=/usr
+          make -j `nproc`
+          sudo make install
+          cd
+      - uses: actions/checkout@v2
+        with:
+          path: ModSecurity-nginx
```
theseion, line 33:
```suggestion
          fetch-depth: 1
```

.github/workflows/ci.yml
```diff
@@ -0,0 +1,65 @@
...
+      - name: Get Nginx source
+        uses: actions/checkout@v3
+        with:
+          repository: nginx/nginx
+          path: nginx
```
theseion, line 38:
```suggestion
          path: nginx
          fetch-depth:
```

.github/workflows/ci.yml
theseion, line 1: Call this file `test.yaml` or something similar. You might later have multiple workflows and `ci.yaml` would be too generic then.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,65 @@
...
+          ./auto/configure --with-ld-opt="-Wl,-rpath,/usr/local/lib" --without-pcre2 --add-module=/home/runner/work/ModSecurity-nginx/ModSecurity-nginx/ModSecurity-nginx
+          make
+          make modules
+          sudo make install
+      - name: Start Nginx
+        run: |
+          sudo /usr/local/nginx/sbin/nginx -c /home/runner/work/ModSecurity-nginx/ModSecurity-nginx/ModSecurity-nginx/.github/nginx.conf
+      - name: Run attack test vhost 1
+        run: |
+          resp=`curl -I -X GET -H "Host: modsectest1" http://localhost/?q=attack | head -1 | cut -d$' ' -f2`
```
theseion, line 51:
```suggestion
          resp=$(curl -sSo /dev/null -w %{http_code} -I -X GET -H "Host: modsectest1" "http://localhost/?q=attack")
```

.github/workflows/ci.yml
```diff
@@ -0,0 +1,65 @@
...
+          if [ $resp == "403" ]; then echo "OK"; else echo "FAIL"; exit 1; fi
```
theseion, line 52:
```suggestion
          if [ "${status}" == "403" ]; then
            echo "OK"
          else
            echo "FAIL"
            exit 1
          fi
```

.github/workflows/ci.yml
```diff
@@ -0,0 +1,65 @@
...
+      - name: Run non-attack test vhost 1
+        run: |
+          resp=`curl -I -X GET -H "Host: modsectest1" http://localhost/?q=1 | head -1 | cut -d$' ' -f2`
```
theseion, line 55:
```suggestion
          set -ex -o pipefail
          status=$(curl -sSo /dev/null -w %{http_code} -I -X GET -H "Host: modsectest1" "http://localhost/?q=1")
```

.github/workflows/ci.yml
```diff
@@ -0,0 +1,65 @@
...
+          if [ $resp == "200" ]; then echo "OK"; else echo "FAIL"; exit 1; fi
```
theseion, line 56:
```suggestion
          if [ "${status}" == "200" ]; then
            echo "OK"
          else
            echo "FAIL"
            exit 1
          fi
```

.github/workflows/ci.yml
```diff
@@ -0,0 +1,65 @@
...
+      - name: Run attack test vhost 2
+        run: |
+          resp=`curl -I -X GET -H "Host: modsectest2" http://localhost/?q=attack | head -1 | cut -d$' ' -f2`
```
theseion, line 59:
```suggestion
          status=$(curl -sSo /dev/null -w %{http_code} -I -X GET -H "Host: modsectest2" "http://localhost/?q=attack")
```

.github/workflows/ci.yml
```diff
@@ -0,0 +1,65 @@
...
+          if [ $resp == "403" ]; then echo "OK"; else echo "FAIL"; exit 1; fi
```
theseion, line 60:
```suggestion
          if [ "${status}" == "403" ]; then
            echo "OK"
          else
            echo "FAIL"
            exit 1
          fi
```

.github/workflows/ci.yml
```diff
@@ -0,0 +1,65 @@
...
+      - name: Run non-attack test vhost 2
+        run: |
+          resp=`curl -I -X GET -H "Host: modsectest2" http://localhost/?q=1 | head -1 | cut -d$' ' -f2`
```
theseion, line 63:
```suggestion
          status=$(curl -sSo /dev/null -w %{http_code} -I -X GET -H "Host: modsectest2" "http://localhost/?q=1")
```

.github/workflows/ci.yml
```diff
@@ -0,0 +1,65 @@
...
+          if [ $resp == "200" ]; then echo "OK"; else echo "FAIL"; exit 1; fi
```
theseion, line 64:
```suggestion
          if [ "${status}" == "200" ]; then
            echo "OK"
          else
            echo "FAIL"
            exit 1
          fi
```

.github/workflows/ci.yml
```diff
@@ -0,0 +1,65 @@
...
+          cd
+      - uses: actions/checkout@v2
+        with:
+          path: ModSecurity-nginx
+      - name: Get Nginx source
+        uses: actions/checkout@v3
+        with:
+          repository: nginx/nginx
+          path: nginx
+      - name: Build nginx
```
theseion, line 39:
```suggestion
      - name: Build nginx with ModSecurity-nginx module
```

theseion (changes requested):

.github/mime.types
airween: Done, see 257f79521187f3f4e1d90ceea87373a5eccfab7b.

.github/workflows/ci.yml
airween: Done, see b52be3a718da26d1464563f3b9ead36f75747ecf and e5877f5ba91c1fc8f185689f58d85462ca245b90.

.github/workflows/ci.yml
airween: Thanks, actually I don't want to use hashes. It would be fine to align each affected repositories (ModSecurity (with both version)), but now those uses the version syntax.

I try to keep it my mind, and that will be the next step. Thanks again.

.github/workflows/ci.yml
airween: Done, see 5aaab8d6c7951fea77f87446930c8c2228e09e95.

.github/workflows/ci.yml
airween: Added `fetch-depth', see a1c9b0082e9472912e091bfc36030bb00e2bd8e5.

.github/workflows/ci.yml
airween: Added `fetch-depth', see a1c9b0082e9472912e091bfc36030bb00e2bd8e5.

.github/workflows/ci.yml
airween: Done, see b5dfe421a0b603d235425ca19f3d3230e461e7b7.

.github/workflows/ci.yml
airween: See my [other commit](https://github.com/owasp-modsecurity/ModSecurity-nginx/pull/317#discussion_r1562428750), same here.

.github/workflows/ci.yml
airween: Done, see 9bad4b0ebd143d9583bce4e3590610ee53bc126c.

.github/workflows/ci.yml
airween: Done, see 9bad4b0ebd143d9583bce4e3590610ee53bc126c.

.github/workflows/ci.yml
airween: Done, see 9bad4b0ebd143d9583bce4e3590610ee53bc126c.

.github/workflows/ci.yml
airween: Done, see 9bad4b0ebd143d9583bce4e3590610ee53bc126c.

.github/workflows/ci.yml
airween: Done, see 9bad4b0ebd143d9583bce4e3590610ee53bc126c.

I didn't put the `set -ex -o piperfail`, because the aim here is to catch the failed request. An `exit 1` is enough here, I guess.

airween:
@theseion many thanks for your suggestion. Please take a review again, if everything is fine I'm going to merge it.

.github/workflows/ci.yml
theseion: That was a remnant of a previous edit, which I had forgotten to remove. I agree, you don't need it.

.github/workflows/test.yml
```diff
@@ -18,25 +18,27 @@ jobs:
...
           sudo apt-get install -y gcc g++ make autoconf automake make libyajl-dev libxml2-dev libmaxminddb-dev libpcre2-dev libpcre2-8-0 libpcre2-16-0 libpcre2-32-0 libcurl4-gnutls-dev jq wget
       - name: Install ModSecurity library
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         run: |
-          modsec=`curl -s https://api.github.com/repos/owasp-modsecurity/ModSecurity/releases/latest | jq .assets[0].browser_download_url | tr -d "\""`
-          modsecver=`curl -s https://api.github.com/repos/owasp-modsecurity/ModSecurity/releases/latest | jq .name | tr -d "\""`
-          wget $modsec -O modsecurity.tar.gz
-          tar -xzf modsecurity.tar.gz
+          modsecver=`gh release view -R owasp-modsecurity/ModSecurity -q .tagName --json tagName`
```
theseion, line 24: You don't need this line. `gh download` fetches the latest release by default.

.github/workflows/test.yml
```diff
@@ -18,25 +18,27 @@ jobs:
...
+          gh release download -p "*.tar.gz" -R owasp-modsecurity/ModSecurity $modsecver -O - | tar -xzf -
```
theseion, line 25:
```suggestion
          gh release download -p "*.tar.gz" -R owasp-modsecurity/ModSecurity -O - | tar -xzf -
```

.github/workflows/test.yml
```diff
@@ -18,25 +18,27 @@ jobs:
...
           cd modsecurity-$modsecver
```
theseion, line 26:
```suggestion
          cd modsecurity-*
```

.github/workflows/test.yml
```diff
@@ -0,0 +1,87 @@
...
+  push:
+  pull_request:
+
+jobs:
+  build-linux:
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [ubuntu-22.04]
+        platform: [x32, x64]
```
theseion, line 13: Currently, your build will run 4 variants (x32 gcc, x32 clang, x64 gcc, x64 clang) but all builds will be identical because you don't make use of `platform` or `compiler`. Comment `platform` and `compiler` until you actually use them.

.github/workflows/test.yml
```diff
@@ -0,0 +1,87 @@
...
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          modsecver=`gh release view -R owasp-modsecurity/ModSecurity -q .tagName --json tagName`
+          gh release download -p "*.tar.gz" -R owasp-modsecurity/ModSecurity $modsecver -O - | tar -xzf -
+          cd modsecurity-$modsecver
+          ./configure --disable-lmdb --prefix=/usr
+          make -j $(nproc)
+          sudo make install
+          cd
+      - uses: actions/checkout@v2
```
theseion, line 31: If you don't want to use hashes, that's your decision, but at least use the latest version of the app:
```suggestion
      - uses: actions/checkout@v4
```

.github/workflows/test.yml
```diff
@@ -0,0 +1,87 @@
...
+        with:
+          path: ModSecurity-nginx
+          fetch-depth: 1
+      - name: Get Nginx source
+        uses: actions/checkout@v3
```
theseion, line 36:
```suggestion
        uses: actions/checkout@v4
```

theseion (changes requested):

.github/workflows/ci.yml
theseion: TBH, I don't see the advantage of aligning the app versions across repositories. I think pinning is much more important.

.github/workflows/test.yml
airween: I know. `$modsecver` needs to know name of the extracted directory in next step.

airween: Done, see 768b388c7f280b3b468f1ab55fdc76bf20c69c1e.

.github/workflows/test.yml
airween: Done, see 768b388c7f280b3b468f1ab55fdc76bf20c69c1e.

.github/workflows/test.yml
airween: Wow, that's really weird it works :smiley:. Thanks, then I remove the `modsecver` variable.

Done, see 768b388c7f280b3b468f1ab55fdc76bf20c69c1e.

.github/workflows/test.yml
airween: Done, see 0c1175463293b0a5720654aec23bbdbcb7cfb8f7.

.github/workflows/test.yml
theseion: This one's still open.

airween: Yes, I'm still looking for the solution. ModSecurity [uses] (almost) same [scheme](https://github.com/owasp-modsecurity/ModSecurity/blob/v3/master/.github/workflows/ci.yml#L11-L14), I tried to copy from there.

airween: Okay, I give up. There is no any difference between platform and compilers - see the [last log](https://github.com/owasp-modsecurity/ModSecurity-nginx/actions/runs/8679760810/job/23799132568) when I tried. Removed those options, we run tests only x64 with gcc.

theseion:
Looks to me like you were trying to set environment variables from scripts, but scripts are isolated, so `export` will have no effect on the next script (neither will a `cd` at the end of a script block btw). What you should do instead is setting the environment variable for the step, like so:

```yaml
      - name: Build nginx with ModSecurity-nginx module
        working-directory: nginx
        env:
          CC: "{{ matrix.compiler }}"
        run: |
          echo "Compiler: ${CC}"

          ./auto/configure --with-ld-opt="-Wl,-rpath,/usr/local/lib" --without-pcre2 --add-module=/home/runner/work/ModSecurity-nginx/ModSecurity-nginx/ModSecurity-nginx
          make
          make modules
          sudo make install
```

airween:
> What you should do instead is setting the environment variable for the step, like so:

Thanks!

How can I use environment variables in evaluation, eg. in an `if` statement?

```yaml
    if [ matrix.compiler == "gcc" ]; then
        ...
```

`${{ matrix.compiler }}` it's not good, because it's the evaluated value (literally `"gcc"`).

theseion:
By passing it through an environment variable like so:

```yaml
- name: a script
  env:
    COMPILER: "${{ matrix.compiler }}"
  run: |
    if [ "${COMPILER}" == "gcc" ]; then
      ...
```

airween:
> By passing it through an environment variable like so:
> 
> ```yaml
> - name: a script
>   env:
>     COMPILER: "${{ matrix.compiler }}"
>   run: |
>     if [ "${COMPILER}" == "gcc" ]; then
>       ...
> ```

I'm afraid in this case the evaluated code will be something similar:
```yaml
    if [ "gcc" == "gcc" ]; then
```
which is always true.

theseion (approved):

airween:
See the newest clear version: #318.

PR #320, opened 2024-04-23
fix: Added missing header for conftest

airween:
In `config` file the `<stdio.h>` header was missing for `printf()` function, which is used in `conftest.c` during `configure` script. This issue was discovered in Debian build system: bug id [1067364](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1067364).

### DIFF
--- a/config
+++ b/config
@@ -10,7 +10,8 @@
 
 ngx_feature_name=
 ngx_feature_run=no
-ngx_feature_incs="#include <modsecurity/modsecurity.h>"
+ngx_feature_incs="#include <modsecurity/modsecurity.h>
+#include <stdio.h>"
 ngx_feature_libs="-lmodsecurity"
 ngx_feature_test='printf("hello");'
 ngx_modsecurity_opt_I=


config
```diff
@@ -10,7 +10,8 @@
 
 ngx_feature_name=
 ngx_feature_run=no
-ngx_feature_incs="#include <modsecurity/modsecurity.h>"
+ngx_feature_incs="#include <modsecurity/modsecurity.h>
+#include <stdio.h>"
```
theseion, line 14: If I were you, I'd add a comment above the line explaining the issue. Otherwise, when someone revisits this in the future, they will not know why the line is there and whether it is safe to remove.

theseion (changes requested):

config
airween: I reviewed few Nginx module during I investigated this issue in Debian, but as I realized there isn't any comment over there. I think the PR's comment is enough here (which describes the issue and the solution too).

Btw, `printf()` needs `<stdio.h>`, that's the "normal" way. So far it has not been a problem because most C compilers added it implicitly.

theseion: Ok, your call.

PR #332, opened 2024-10-22
Do not allocate new strings, add support for HTTP3

JakubOnderka:
Add support for HTTP3 protocol, that is available in newer nginx.

And also avoid allocating new string for common HTTP methods to make modsecurity little bit faster and consume less memory.

### DIFF
--- a/src/ngx_http_modsecurity_header_filter.c
+++ b/src/ngx_http_modsecurity_header_filter.c
@@ -471,7 +471,7 @@ ngx_http_modsecurity_header_filter(ngx_http_request_t *r)
      */
     for (i = 0; ngx_http_modsecurity_headers_out[i].name.len; i++)
     {
-        dd(" Sending header to ModSecurity - header: `%.*s'.",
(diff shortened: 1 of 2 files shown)


airween:
Hi @JakubOnderka,

first of all, many thanks for your contribution!

Meanwhile we added some new test to our CI workflow. It would be fine to pick up those modifications and runt those tests. Also would be fine to make some new tests to check the HTTP3 behavior.

Let us know if you have any question or we can help you anything. Also, you can join to our [Slack](https://owasp.org/slack/invite) server (look for the [#project-modsecurity](https://owasp.slack.com/archives/C069PCXSW12) channel).

fzipi:
@JakubOnderka can you split this in two PRs? One for HTTP3, an dthe other for the allocation? I think it will help merging them separately...

JakubOnderka:
Closing, HTTP3 support is added in #364 and I moved avoiding allocation to #367

PR #342, opened 2025-02-13
Add a Directive to get a Proxied Client's IP Address (Closes #341)

### LINKED ISSUE #341: Add an Option to Log Proxied Client IPs
When using the Proxy Protocol, the ModSecurity module isn't currently able to access the client's IP. Thus, only the proxying server's IP is available.

For a project I work on, we needed an option to do so and implemented a patch we would like to upstream. Could I open a PR for it?

thekief:
As written in #341, this patch adds functionality to obtain a client's IP address, if the request has been proxied using the Proxy Protocol.

Besides the check, if the option is enabled, I added a check if the Proxy Protocol is actually being used.

### DIFF
--- a/src/ngx_http_modsecurity_rewrite.c
+++ b/src/ngx_http_modsecurity_rewrite.c
@@ -78,10 +78,23 @@ ngx_http_modsecurity_rewrite_handler(ngx_http_request_t *r)
          * erliest phase that nginx allow us to attach those kind of hooks.
          *
          */
-        int client_port = ngx_inet_get_port(connection->sockaddr);
+        int client_port;
+
+        if (mcf->proxy_protocol_ip && connection->proxy_protocol) {
+          client_port = connection->proxy_protocol->src_port;
+        } else {
+          client_port = ngx_inet_get_port(connection->sockaddr);
+        }
         int server_port = ngx_inet_get_port(connection->local_sockaddr);
 
-        const char *client_addr = ngx_str_to_char(addr_text, r->pool);
+        const char *client_addr;
+
+        if (mcf->proxy_protocol_ip && connection->proxy_protocol) {
+          client_addr =  ngx_str_to_char(connection->proxy_protocol->src_addr, r->pool);
+        } else {
+          client_addr = ngx_str_to_char(addr_text, r->pool);
+        }
+
         if (client_addr == (char*)-1) {
             return NGX_HTTP_INTERNAL_SERVER_ERROR;
         }
--- a/README.md
+++ b/README.md
@@ -5,7 +5,7 @@
 [![](https://raw.githubusercontent.com/ZenHubIO/support/master/zenhub-badge.png)](https://zenhub.com)
 
 
-The ModSecurity-nginx connector is the connection point between nginx and libmodsecurity (ModSecurity v3). Said another way, this project provides a communication channel between nginx and libmodsecurity. This connector is required to use LibModSecurity with nginx. 
+The ModSecurity-nginx connector is the connection point between nginx and libmodsecurity (ModSecurity v3). Said another way, this project provides a communication channel between nginx and libmodsecurity. This connector is required to use LibModSecurity with nginx.
 
 The ModSecurity-nginx connector takes the form of an nginx module. The module simply serves as a layer of communication between nginx and ModSecurity.
 
@@ -176,6 +176,17 @@ using the same unique identificator.
 
 String can contain variables.
 
+modsecurity_proxy_protocol_ip
+-----------
+**syntax:** *modsecurity_proxy_protocol_ip on | off*
+
+**context:** *http, server, location*
+
+**default:** *off*
+
+If activated, ModSecurity logs the actual IP address of a client when the connections is proxied using the Proxy Protocol.
+If a connection is proxied without this directive activated, the logged IP address is the one of the proxy server.
+
 
 # Contributing
 
@@ -194,7 +205,7 @@ here: https://help.github.com/articles/using-pull-requests/
 Please respect the coding style in use. Pull requests can include various commits, so
 provide one fix or one functionality per commit. Do not change anything outside
 the scope of your target work (e.g. coding style in a function that you have
-passed by). 
+passed by).
 
 ### Don’t know where to start?
 
@@ -211,9 +222,9 @@ You may also take a look at recent bug reports and open issues to get an idea of
 ### Testing your patch
 
 Along with the manual testing, we strongly recommend that you to use the nginx test
-utility to make sure that you patch does not adversely affect the behavior or performance of nginx. 
+utility to make sure that you patch does not adversely affect the behavior or performance of nginx.
 
-The nginx tests are available on: http://hg.nginx.org/nginx-tests/ 
+The nginx tests are available on: http://hg.nginx.org/nginx-tests/
 
 To use those tests, make sure you have the Perl utility prove (part of Perl 5)
 and proceed with the following commands:
@@ -226,7 +237,7 @@ $ TEST_NGINX_BINARY=/path/to/your/nginx prove .
 
 If you are facing problems getting your added functionality to pass all the nginx tests, feel free to contact us or the nginx mailing list at: http://nginx.org/en/support.html
 
-### Debugging 
+### Debugging
 
 We respect the nginx debugging schema. By using the configuration option
 "--with-debug" during the nginx configuration you will also be enabling the
--- a/src/ngx_http_modsecurity_common.h
+++ b/src/ngx_http_modsecurity_common.h
@@ -126,6 +126,7 @@ typedef struct {
 #endif
 
     ngx_http_complex_value_t  *transaction_id;
+    ngx_flag_t                 proxy_protocol_ip;
 } ngx_http_modsecurity_conf_t;
 
 
--- a/src/ngx_http_modsecurity_module.c
+++ b/src/ngx_http_modsecurity_module.c
@@ -514,6 +514,14 @@ static ngx_command_t ngx_http_modsecurity_commands[] =  {
     0,
     NULL
   },
+  {
+    ngx_string("modsecurity_proxy_protocol_ip"),
+    NGX_HTTP_LOC_CONF|NGX_HTTP_SRV_CONF|NGX_HTTP_MAIN_CONF|NGX_CONF_FLAG,
+    ngx_conf_set_flag_slot,
+    NGX_HTTP_LOC_CONF_OFFSET,
+    offsetof(ngx_http_modsecurity_conf_t, proxy_protocol_ip),
+    NULL
+  },
   ngx_null_command
 };
 
@@ -725,6 +733,7 @@ ngx_http_modsecurity_create_conf(ngx_conf_t *cf)
     conf->rules_set = msc_create_rules_set();
     conf->pool = cf->pool;
     conf->transaction_id = NGX_CONF_UNSET_PTR;
+    conf->proxy_protocol_ip = NGX_CONF_UNSET;
 #if defined(MODSECURITY_SANITY_CHECKS) && (MODSECURITY_SANITY_CHECKS)
     conf->sanity_checks_enabled = NGX_CONF_UNSET;
 #endif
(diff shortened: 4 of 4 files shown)


thekief:
Please let me know, if the whitespace changes in the readme should be removed. I made the changes using VS Codium and didn't know that it forces a newline encoding.

airween:
> Please let me know, if the whitespace changes in the readme should be removed.

No worries, that's in a good place.

Anyway, thanks for the PR - I'm thinking about we should add new test cases for each new PR (which adds a new feature - like this). I mean please take a look at the [regression test](https://github.com/owasp-modsecurity/ModSecurity-nginx/blob/master/.github/workflows/test.yml) file of this repository, how could we make a config where we can see that the modified code does what the author wants. And this is important, because if someone in the future sends a new PR, we must be sure this behavior does not change. Do you have any idea?

thekief:
Yeah, I have an idea for a test case. `curl` not only supports the Proxy Protocol [^1], but also the option to set the client's IP address [^2]. 

If that's alright with you, I would add 2 new endpoints to the test configuration:
- one that returns the client's ip with the new directive enabled
- another one hat returns the client's ip with the new directive disabled

What do you think of that? 

[^1]: https://curl.se/docs/manpage.html#:~:text=also%20-x%2C%20--proxy.-,--haproxy-protocol,-(HTTP)%20Send%20a
[^2]: https://curl.se/docs/manpage.html#:~:text=--haproxy-clientip%20%3Cip%3E

airween:
Hi @thekief,

I tried to check your patch, but may be I'm doing something wrong, I can't see the client's IP in my log.

Here is my config:
```
server {
    listen 8088;
    server_name _;

    root /var/www/html;
    index index.html;

    modsecurity on;
    modsecurity_rules_file /etc/nginx/modsecurity_includes.conf;
    modsecurity_proxy_protocol_ip on;

    error_log /var/log/nginx/backend_error.log info;
    access_log /var/log/nginx/backend_acces.log;

    location / {
        try_files $uri $uri/ =404;
    }
}

server {
    listen          8080;
    server_name     proxytest;

    modsecurity off;
    location / {
        proxy_read_timeout 300;
        proxy_connect_timeout 300;
        proxy_set_header X-Real-IP $remote_addr;
        proxy_set_header X-Forwarded-For $remote_addr;
        proxy_set_header Host $host;
        proxy_set_header X-Forwarded-Proto $scheme;
        proxy_pass http://127.0.0.1:8088;
        proxy_http_version 1.1;
        proxy_set_header Upgrade $http_upgrade;
        proxy_buffering off;
        proxy_redirect http:// https://;
    }
}
```
As you can see I created a simple vhost for static files, there I turned on modsecurity, load rules and turned on your new feature. I also set a separated logs (both access and error).

Below that I created a proxy which sends all requests to that site.

Here is how I checked that:

```
curl --interface 172.35.40.12 -H "Host: proxytest" "http://localhost:8080/?q=/bin/bash"
```

where `172.35.40.12` is the IP of my physical interface, but the request sent to localhost (with that IP) with HTTP header `Host: proxytest` (see above the proxy config).

With this request I get this line in my error.log:
```
2025/02/19 21:04:29 [info] 255143#255143: *3 ModSecurity: Warning. Matched "Operator `ValidateByteRange' with parameter `38,44-46,48-58,61,65-90,95,97-122' against variable `ARGS:q' (Value: `/bin/bash' ) [file "/home/airween/src/coreruleset/rules/REQUEST-920-PROTOCOL-ENFORCEMENT.conf"] [line "1723"] [id "920273"] [rev ""] [msg "Invalid character in request (outside of very strict set)"] [data "ARGS:q=/bin/bash"] [severity "2"] [ver "OWASP_CRS/4.12.0-dev"] [maturity "0"] [accuracy "0"] [tag "application-multi"] [tag "language-multi"] [tag "platform-multi"] [tag "attack-protocol"] [tag "paranoia-level/4"] [tag "OWASP_CRS"] [tag "capec/1000/210/272"] [hostname "127.0.0.1"] [uri "/"] [unique_id "173999546951.849862"] [ref "o0,1o4,1v8,9t:urlDecodeUni"], client: 127.0.0.1, server: _, request: "GET /?q=/bin/bash HTTP/1.1", host: "proxytest"
```

Where should I see my client IP instead of `127.0.0.1` (which is the proxy's IP)?

Or may be I misunderstand something.

thekief:
To clear up a small misunderstanding, the "Proxy Protocol" I am talking about, is specified in the `listen`-directive [^1].

I will add an example configuration tomorrow.

[^1]: https://nginx.org/en/docs/http/ngx_http_core_module.html#:~:text=%20the%20proxy_protocol%20parameter%20(1.5.12)%20allows%20specifying%20that%20all%20connections%20accepted%20on%20this%20port%20should%20use%20the%20proxy%20protocol.%20

airween:
> To clear up a small misunderstanding, the "Proxy Protocol" I am talking about, is specified in the `listen`-directive [1](#user-content-fn-1-cdc86d64bce4d0d3248e3fbc7707ea69).
> 
> I will add an example configuration tomorrow.

Thanks - then I can help you to create test cases. And of course, examples are always helpful.

airween:
@thekief any update on this?

thekief:
Hi,
I'm really sorry for my late reply, I had to deal with other things -.-

> Thanks - then I can help you to create test cases. And of course, examples are always helpful.
As for the test, I would propose the following:

`curl` has the following two options in regard to the Proxy Protocol:

```
--haproxy-clientip <ip>                       Set address in HAProxy PROXY
--haproxy-protocol                            Send HAProxy PROXY protocol v1 header
```

So, for example we could use `curl --haproxy-protocol --haproxy-clientip 123.123.123.123 localhost:8443` as the test command.

If `modsecurity_proxy_protocol_ip on;` is set and a rule is triggered, `123.123.123.123` should be visible in the error log. If `modsecurity_proxy_protocol_ip off;` is set `127.0.0.1` or `localhost` should be visible in the log files.

thekief:
Regarding configuration, I would have done the following:

**nginx.conf**:
```nginx
pid /tmp/nginx.pid;

error_log  /dev/stdout  info;

events {
    worker_connections   2000;
}

user www-data;

http {
    server {
        listen 8080 proxy_protocol;

        root /var/www/html;
        index index.html;

        modsecurity on;
        modsecurity_rules_file /etc/nginx/modsecurity_includes.conf;
        modsecurity_proxy_protocol_ip off;

        location / {
            return 200;
        }
    }

    server {
        listen 8081 proxy_protocol;

        root /var/www/html;
        index index.html;

        modsecurity on;
        modsecurity_rules_file /etc/nginx/modsecurity_includes.conf;
        modsecurity_proxy_protocol_ip on;

        location / {
            return 200;
        }
    }
}
```

**REQUEST-900-EXCLUSION-RULES-BEFORE-CRS.conf**: 
```
SecRuleRemoveByTag  OWASP_CRS

SecRule REQUEST_HEADERS "@contains worm" \
        "id:1012 \
        ,phase:1 \
        ,deny \
        ,t:none \
        ,status:500 \
        ,log \
        "
```

Resulting in the following output when executing `curl --haproxy-clientip 123.123.123.123 --haproxy-protocol -v -k -H "Animal: worm" localhost:8080`:

```json
{"transaction":{"client_ip":"172.21.0.1","time_stamp":"Wed Jun 11 13:32:17 2025","server_id":"575f816b3c1769ef25a49fc7dcabe0a096d605ee","client_port":37068,"host_ip":"172.21.0.2","host_port":8080,"unique_id":"17496487374.770509","request":{"method":"GET","http_version":1.1,"uri":"/","headers":{"Host":"localhost:8080","User-Agent":"curl/8.14.1","Accept":"*/*","Animal":"worm"}}...
```

If `curl --haproxy-clientip 123.123.123.123 --haproxy-protocol -v -k -H "Animal: worm" localhost:8081` is executed, the following output is in the logs:

```json
{"transaction":{"client_ip":"123.123.123.123","time_stamp":"Wed Jun 11 13:32:12 2025","server_id":"575f816b3c1769ef25a49fc7dcabe0a096d605ee","client_port":42630,"host_ip":"172.21.0.2","host_port":8081,"unique_id":"174964873292.593363","request":{"method":"GET","http_version":1.1,"uri":"/","headers":{"Host":"localhost:8081","User-Agent":"curl/8.14.1","Accept":"*/*","Animal":"worm"}}...
```

airween:
> Resulting in the following output when executing `curl --haproxy-clientip 123.123.123.123 --haproxy-protocol -v -k -H "Animal: worm" localhost:8080`:

with this request I got a `400` response:

```
curl --haproxy-clientip 123.123.123.123 --haproxy-protocol -v "http://my.host:8080/?q=/bin/bash"
* Host my.host:8080 was resolved.
* IPv6: (none)
* IPv4: 192.168.72.16
*   Trying 192.168.72.16:8080...
* Established connection to my.host (192.168.72.16 port 8080) from 192.168.72.16 port 40546 
* using HTTP/1.x
> GET /?q=/bin/bash HTTP/1.1
> Host: my.host:8080
> User-Agent: curl/8.17.0
> Accept: */*
> 
* Request completely sent off
< HTTP/1.1 400 Bad Request
< Server: nginx/1.28.0
< Date: Mon, 08 Dec 2025 15:00:57 GMT
< Content-Type: text/html
< Content-Length: 157
< Connection: close
```

thekief:
Sorry for the late reply. Did you check the ModSecurity logs? They are relevant, as I could already previously modify the nginx configuration as needed.

Please retest it and let me know.

src/ngx_http_modsecurity_rewrite.c
```diff
@@ -78,10 +78,23 @@ ngx_http_modsecurity_rewrite_handler(ngx_http_request_t *r)
          * erliest phase that nginx allow us to attach those kind of hooks.
          *
          */
-        int client_port = ngx_inet_get_port(connection->sockaddr);
+        int client_port;
+
+        if (mcf->proxy_protocol_ip && connection->proxy_protocol) {
```
airween: I'm almost sure there is a very small chance, but `connection->proxy_protocol` exists after 1.17 (1.17.6). Would you mind to add a macro as a guard to avoid compilation error on older Nginx's?

src/ngx_http_modsecurity_rewrite.c
```diff
@@ -78,10 +78,23 @@ ngx_http_modsecurity_rewrite_handler(ngx_http_request_t *r)
          * erliest phase that nginx allow us to attach those kind of hooks.
          *
          */
-        int client_port = ngx_inet_get_port(connection->sockaddr);
+        int client_port;
+
```
airween:
```suggestion
#if defined(nginx_version) && (nginx_version >= 1017006)
```
I'm almost sure there is a very small chance, but `connection->proxy_protocol` exists after 1.17 (1.17.6). Would you mind to add a macro as a guard to avoid compilation error on older Nginx's?

airween, line 82: I'm almost sure there is a very small chance, but `connection->proxy_protocol` exists after 1.17 (1.17.6). Would you mind to add a macro as a guard to avoid compilation error on older Nginx's?

Something like
```C
#if defined(nginx_version) && nginx_version >= 1017006
        if (mcf->proxy_protocol_ip && connection->proxy_protocol) {
          client_port = connection->proxy_protocol->src_port;
        } else {
          client_port = ngx_inet_get_port(connection->sockaddr);
        }
#else
      client_port = ngx_inet_get_port(connection->sockaddr);
#endif
```

airween:
Hi @thekief,

sorry for the late reply - I try to process all PR's now.

Please take a look at my note above and Copilot's suggestions.

Also:
* what if the admin set `modsecurity_proxy_protocol_ip on;` but forget the `listen ... proxy_protocol`? May be the module will log the proxy IP into the logfile? Wouldn't be add some debug info?
* Could you add some detailed information about this change? I mean explain why is this change (and using of the given directive) is better that use `realip` module? (I don't use any of them, just tried to review available solutions and found that module). To me for the reason is enough that with this solution nobody needs to use `realip` module
* could you add some test (as Copilot mentioned above too)?

Issue #258, opened 2021-12-15
Match rule transform POST in GET 


cello86:
Hi All,
we noticed on  modsecurity for Nginx connector a strange behavior. If a rule is matched before the normal CRS rules engine the HTTP method is noticed like GET instead the original POST. 

We already had some issues with the internal redirects but the matching on different phases can generate this type of error?

Thanks,
Marcello

martinhsv:
Hello @cello86 ,

I'm not sure I understand the use case that you are describing.

Note that ModSecurity (including the nginx connector) is the "engine".  CRS is a generic rule set that is often used with ModSecurity.

From what have included, however, what you are seeing sounds like it might be expected behaviour related to error redirects. You may wish to have a look at the following for more detail: https://github.com/SpiderLabs/ModSecurity-nginx/issues/152#issuecomment-583037424

If you think what you are experiencing is unrelated to that, please provide additional detail.

cello86:
Hi @cello86,
we configured the nginx instance with the error page without mod_security enabled to avoid the internal redirect.

```
    server {
	server_name test.local;

        modsecurity on;
        modsecurity_rules_file /usr/local/nginx/conf/modsecurity.conf;
 
        error_page 403 /error/403.json;

        location ^~ /error {
            modsecurity_rules 'SecRule REQUEST_URI "@beginsWith /" "id:1,pass,phase:2,log,ctl:ruleEngine=DetectionOnly"';
            alias "/usr/local/nginx/error";
        }

 
        location / {
            root   html;
        }
    }
```

Marcello

martinhsv:
Hello @cello86 ,

Perhaps I've misunderstood something but how does that configuration result in 'without mod_security enabled'?

You explicitly have 'modsecurity on' within the outside server block. So it will still be 'on' inside the 'location ^~ /error' -- which it seems you understand because you include an extra modsecurity_rules entry there.

That extra rule includes an action 'ctl:ruleEngine=DetectionOnly', so rules that are impacted by that will, of course, use Detection Only.

However even for rule detections after that point, you will still see the effect of the GET requests that were originally POSTs (e.g. in logging). Note also that your "id:1" rule will run later than at least some other rules (you have it as a phase:2 rule), you could have a look at the DebugLog output to confirm which rules are being run in which order.

If none of the above clarifies things for you, I think you're going to have to describe in more detail, what effects exactly you are seeing and how you expect them to be different.

martinhsv:
[...]

PR #346, opened 2025-02-19
fix: recovery context after internal redirect, re-add #273

airween:
This PR is a "copy" of #273, plus I added some tests to check the behavior that the patch adds.

Original credits to @liudongmiao.

@liudongmiao could you take a review?

### DIFF
--- a/.github/nginx/docs/403.html
+++ b/.github/nginx/docs/403.html
@@ -1,0 +1,10 @@
+<html>
+<head>
+<title>403</title>
+</head>
+
+<body>
+Forbidden 403 - custom error page.
+</body>
+</html>
+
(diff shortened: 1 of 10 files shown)


liudongmiao:
@airween what's the difference with https://github.com/owasp-modsecurity/ModSecurity-nginx/pull/334?

airween:
> @airween what's the difference with #334?

Nothing. As I explained in my [initial](https://github.com/owasp-modsecurity/ModSecurity-nginx/pull/346#issue-2863644259) comment, this is a copy of #344.

But if you take a look at the [history](https://github.com/owasp-modsecurity/ModSecurity-nginx/commits/master/), I had to revert last two PR's. Then I added few additional tests, and now I would like to re-add your fix again.

So I hope it's a copy of #344 :), this is what would be a help to review.

liudongmiao:
@airween It seems almost same.

However, I'd suggest you check whether it solves #182, #255, #258 like original PR.

Mik13:
+1

airween:
Ok, I'm going to merge this PR, let's see how it works.

PR #353, opened 2025-04-27
feat: set correct hostname in log produced by Nginx

airween:
**Reason**

There was a discussion about this feature: [#ModSecurity/3200](https://github.com/owasp-modsecurity/ModSecurity/issues/3200)

Current behavior: if ModSecurity catches an attack then it produces log entries. The problem is that the `[hostname]` field contains the server's IP address - which carries no information at all:

```
ModSecurity: Warning. ... [hostname "18.19.20.21"] [uri "/xmlrpc.php"]... client: 91.92.93.94, server: www.myserver.com, request: "POST /xmlrpc.php HTTP/1.1", host: "www.myserver.com"
```
At the end of the line, Nginx (and not ModSecurity) puts other fields, like `server` and `host`, but unfortunately those can be truncated if the other parts of line are too long, eg. `[data]` field by ModSecurity, or `request` field by Nginx, because Nginx truncates the log line after [2048](https://github.com/nginx/nginx/blob/master/src/core/ngx_log.h#L76) bytes.

The other advantage of this patch that now the fields will be the same as in case of mod_security2, so parsing the lines (hopefully) will be easier.

Note, that previously I sent a [PR](https://github.com/owasp-modsecurity/ModSecurity/pull/2906) to ModSecurity, but it was rejected. Now the result is the same but the solution is completely different, because the library gets the value from the HTTP server.

**How does it work**

In [ngx_http_modsecurity_rewrite.c](https://github.com/airween/ModSecurity-nginx/blob/sethostname/src/ngx_http_modsecurity_rewrite.c#L89-L125) there is a new block (with a compilation condition - it needs libmodsecurity3 3.0.13 at least, because [msc_set_request_hostname](https://github.com/owasp-modsecurity/ModSecurity/blob/v3/master/src/transaction.cc#L2316-L2319) was implemented in that version). In this block the module try to get the `Host`  header from the request which processed by Nginx. If there is no `Host` header (eg. in case of HTTP/1.0) or the length is 0, then it gets the `server_name` value from the used virtualhost context - note, that it can be "`_`" in default config. Also if there is no `Host` header, Nginx will apply the default vhost, so the line will looks like `[hostname "_"]`.

In other cases (there **is** a `Host` header) that value will be in `[hostname]` field.

### DIFF
--- a/src/ngx_http_modsecurity_rewrite.c
+++ b/src/ngx_http_modsecurity_rewrite.c
@@ -86,6 +86,44 @@ ngx_http_modsecurity_rewrite_handler(ngx_http_request_t *r)
             return NGX_HTTP_INTERNAL_SERVER_ERROR;
         }
 
+#if defined(MODSECURITY_CHECK_VERSION)
+#if MODSECURITY_VERSION_NUM >= 30130100
+        ngx_str_t hostname;
+        // first check if Nginx received a Host header and it's usable
+        // (i.e. not empty)
+        // if yes, we can use that
+        if (r->headers_in.server.len > 0) {
+            hostname.len = r->headers_in.server.len;
+            hostname.data = r->headers_in.server.data;
+        }
+        else {
+            // otherwise we try to use the server config, namely the
+            // server_name $SERVER_NAME
+            // directive
+            // for eg. in default config, server_name is "_"
+            // possible all requests without a Host header will be
+            // handled by this server block
+            ngx_http_core_srv_conf_t  *cscf;
+            cscf = ngx_http_get_module_srv_conf(r, ngx_http_core_module);
+            if (cscf->server_name.len > 0) {
+                hostname.len = cscf->server_name.len;
+                hostname.data = cscf->server_name.data;
+            }
+        }
+        if (hostname.len > 0) {
+            const char *host_name = ngx_str_to_char(hostname, r->pool);
+            if (host_name == (char*)-1 || host_name == NULL) {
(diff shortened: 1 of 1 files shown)


dune73:
Could give us three times the same alert in ModSec2, ModSec3-without-patch and ModSec3-with-patch? That would be easier to understand.

airween:
Sure.

The request:

```
curl -v -H "Host: foobar.com" "http://172.35.40.12:80/?q=/bin/bash"
```

Apache2 log:

```
ModSecurity: Warning. .... [file "...REQUEST-932-APPLICATION-ATTACK-RCE.conf"] [line "626"] [id "932160"] ... [tag "application-multi"] ... [tag "PCI/6.5.2"] [hostname "foobar.com"] [uri "/"]
```
Libmodsecurity3 + Nginx **without** patch:
```
ModSecurity: Warning. ... [file "...REQUEST-932-APPLICATION-ATTACK-RCE.conf"] [line "606"] [id "932160"] ... [tag "application-multi"] ... [tag "PCI/6.5.2"] [hostname "172.35.40.12"] [uri "/"]
```
Libmodsecurity3 + Nginx **with** patch:
```
ModSecurity: Warning. ... [file "...REQUEST-932-APPLICATION-ATTACK-RCE.conf"] [line "606"] [id "932160"] ... [tag "application-multi"] ... [tag "PCI/6.5.2"] [hostname "foobar.com"] [uri "/"]
```

airween:
No other remarks, I'm going to merge this modification.

Issue #206, opened 2020-05-28
[Perf] Don't read full response into mem if not needed NGINX


jeremyjpj0916:
Regarding this bit here:

https://github.com/SpiderLabs/ModSecurity-nginx/blob/master/src/ngx_http_modsecurity_header_filter.c#L450

Got me thinking, for users that just want to waf protect inbound client requests we really can drop all the phases that execute after the ```proxy_pass``` directive(besides the log phase of course) if we are talking nginx as a gateway or lb. So I did something like this in code:

Basically removed all references to _body and _header filter logic

config
```
# vim: filetype=sh

# If $NGX_IGNORE_RPATH is set to "YES", we will ignore explicit
# library path specification on resulting binary, allowing libmodsecurity.so
# to be relocated across configured library pathes (adjust /etc/ld.so.conf
# or set $LD_LIBRARY_PATH environment variable to manage them)
#
# $YAJL_LIB variable may need to be populated in case of non-standard
# path of libyajl.so's installation

ngx_feature_name=
ngx_feature_run=no
ngx_feature_incs="#include <modsecurity/modsecurity.h>"
ngx_feature_libs="-lmodsecurity"
ngx_feature_test='printf("hello");'
ngx_modsecurity_opt_I=
ngx_modsecurity_opt_L=

YAJL_EXTRA=
if test -n "$YAJL_LIB"; then
    YAJL_EXTRA="-L$YAJL_LIB -lyajl"
fi

# If $MODSECURITY_INC is specified, lets use it. Otherwise lets try
# the default paths
#
if [ -n "$MODSECURITY_INC" -o -n "$MODSECURITY_LIB" ]; then
    # explicitly set ModSecurity lib path
    ngx_feature="ModSecurity library in \"$MODSECURITY_LIB\" and \"$MODSECURITY_INC\" (specified by the MODSECURITY_LIB and MODSECURITY_INC env)"
    ngx_feature_path="$MODSECURITY_INC"
    ngx_modsecurity_opt_I="-I$MODSECURITY_INC"
    ngx_modsecurity_opt_L="-L$MODSECURITY_LIB $YAJL_EXTRA"

    if [ $NGX_RPATH = YES ]; then
        ngx_feature_libs="-R$MODSECURITY_LIB -L$MODSECURITY_LIB -lmodsecurity $YAJL_EXTRA"
[...]

PR #356, opened 2025-06-03
Add an Option to Skip Body Inspections ( Closes #343 )

### LINKED ISSUE #343: Add an Option to Skip Body Inspections
I have a rather rare use-case, where sometimes large POST requests are made to ModSecurity protected endpoints. The sent data itself is encrypted and cannot be inspected. As ModSecurity buffers the whole requests, this leads to quite a big memory consumption, leading to the container being killed, due to the memory limits.

To address this issue, we developed a patch that allows on a per-location basis to skip the body inspection, while keeping the rest of the ModSecurity checks. I'd like to open a pull request to evaluate, if that patch fits into the official source.

thekief:
As dicussed in the issue, I would like to add 2 new directives that allow to skip the body inspection. There are a few usecases, where, e.g. encrypted data is set, and no useful inspection can be made.

While denying the body access may skip the inspection, it still results in the caching of data. Subsequently, there is an unneeded amount of resource consumption, memory, as well as time, involved. 

While it would be cleaner to expose an API that allows users of `libmodsecurity` to check, if a path, e.g. has a reqeust body check, this involves a lot more work. The approach taken in this PR, shifts the functionality to that is only necessitates changes in the `nginx` module.

**Regarding Tests:** As for tests, I'm a bit unsure what the best way would be to test it. One way would be to try to upload a file, e.g. 100MB, to a location and check when the first data reaches it. If the body inspection is disabled, the first bytes will arrive much sooner, as the `nginx` module caches the body until it's fully received otherwise and only then forwards the data.

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,7 +45,7 @@ Further information about nginx third-party add-ons support are available [here]
 # Usage
 
 ModSecurity for nginx extends your nginx configuration directives.
-It adds four new directives and they are:
+It adds six new directives and they are:
 
 modsecurity
 -----------
@@ -191,6 +191,60 @@ As an open source project we invite (and encourage) anyone from the community to
 functionality, bug fixes, bug reports, beginners user support, and anything else that you
 are willing to help with. Thank you.
 
+modsecurity_skip_req_body_filter
+-----------------
+**syntax:** *modsecurity_skip_req_body_filter on | off*
+
+**context:** *http, server, location*
+
+**default:** *off*
+
+Allows to skip the caching of the request body and subsequently its inspection.
+Useful in cases, where `SecRequestBodyAccess` or `ctl:requestBodyAccess` is set, due to, e.g. encrypted data, as the caching causes an unneeded memory overhead.
+
+
+```nginx
+server {
+    modsecurity on;
+    modsecurity_rules_file /etc/my_modsecurity_rules.conf;
+
+    location / {
+        root /var/www/html;
+    }
+
+    location = /special/unchecked/path {
+        # skip the inspection of the request body
+        modsecurity_skip_req_body_filter on;
+    }
+}
+```
+
+modsecurity_skip_resp_body_filter
+-----------------
+**syntax:** *modsecurity_skip_resp_body_filter on | off*
+
+**context:** *http, server, location*
+
+**default:** *off*
+
+Allows to skip the caching of the request body and subsequently its inspection.
+Useful in cases, where `SecResponseBodyAccess` is set, due to, e.g. encrypted data, as the caching causes an unneeded memory overhead.
+
+```nginx
+server {
+    modsecurity on;
+    modsecurity_rules_file /etc/my_modsecurity_rules.conf;
+
+    location / {
+        root /var/www/html;
+    }
+
+    location = /special/unchecked/path {
+        # skip the inspection of the response body
+        modsecurity_skip_resp_body_filter on;
+    }
+}
+```
 
 ## Providing Patches
 
--- a/README.md
+++ b/README.md
@@ -45,7 +45,7 @@ Further information about nginx third-party add-ons support are available [here]
 # Usage
 
 ModSecurity for nginx extends your nginx configuration directives.
-It adds six new directives and they are:
+It adds seven new directives and they are:
 
 modsecurity
 -----------
--- a/src/ngx_http_modsecurity_access.c
+++ b/src/ngx_http_modsecurity_access.c
@@ -56,6 +56,11 @@ ngx_http_modsecurity_access_handler(ngx_http_request_t *r)
         return NGX_DECLINED;
     }
 
+    if(mcf->skip_req_body_filter == 1) {
+        dd("Skipping request body filter");
+        return NGX_DECLINED;
+    }
+
     /*
     if (r->method != NGX_HTTP_GET &&
         r->method != NGX_HTTP_POST && r->method != NGX_HTTP_HEAD) {
--- a/src/ngx_http_modsecurity_body_filter.c
+++ b/src/ngx_http_modsecurity_body_filter.c
@@ -39,8 +39,8 @@ ngx_http_modsecurity_body_filter(ngx_http_request_t *r, ngx_chain_t *in)
 {
     ngx_chain_t *chain = in;
     ngx_http_modsecurity_ctx_t *ctx = NULL;
+    ngx_http_modsecurity_conf_t *mcf = NULL;
 #if defined(MODSECURITY_SANITY_CHECKS) && (MODSECURITY_SANITY_CHECKS)
-    ngx_http_modsecurity_conf_t *mcf;
     ngx_list_part_t *part = &r->headers_out.headers.part;
     ngx_table_elt_t *data = part->elts;
     ngx_uint_t i = 0;
@@ -50,7 +50,19 @@ ngx_http_modsecurity_body_filter(ngx_http_request_t *r, ngx_chain_t *in)
         return ngx_http_next_body_filter(r, in);
     }
 
-    ctx = ngx_http_modsecurity_get_module_ctx(r);
+    mcf = ngx_http_get_module_loc_conf(r, ngx_http_modsecurity_module);
+
+    if (mcf == NULL){
+        dd("failed to get configuration");
+        return NGX_HTTP_INTERNAL_SERVER_ERROR;
+    }
+
+    if (mcf->skip_resp_body_filter) {
+        dd("Skipping response body filter");
+        return ngx_http_next_body_filter(r, in);
+    }
+
+    ctx = ngx_http_get_module_ctx(r, ngx_http_modsecurity_module);
 
     dd("body filter, recovering ctx: %p", ctx);
 
@@ -63,8 +75,7 @@ ngx_http_modsecurity_body_filter(ngx_http_request_t *r, ngx_chain_t *in)
     }
 
 #if defined(MODSECURITY_SANITY_CHECKS) && (MODSECURITY_SANITY_CHECKS)
-    mcf = ngx_http_get_module_loc_conf(r, ngx_http_modsecurity_module);
-    if (mcf != NULL && mcf->sanity_checks_enabled != NGX_CONF_UNSET)
+    if (mcf->sanity_checks_enabled != NGX_CONF_UNSET)
     {
 #if 0
         dd("dumping stored ctx headers");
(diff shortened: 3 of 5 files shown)


airween:
Hi @thekief,

thanks for this PR, I think this is quite useful.

> **Regarding Tests:** As for tests, I'm a bit unsure what the best way would be to test it.

If I suggest suggest a method:

* create a vhost config where you turn `Off` the request body inspection
* create a rule (into `phase:2`) that checks for a variable from the request body (eg `ARGS_POST`)
* send a request which should trigger that rule

For the response body skip:

* create another vhost config where you turn `Off` the response body inspection
* create a rule (into `phase:4`) that checks for a variable from the response body (eg `RESPONSE_BODY`)
* set a content for that vhost which triggers that rule

You can add cross checking too (eg. the first mentioned rule above executes in `phase:4` and you try to mix the settings with these new options...)

thekief:
Sorry for getting back so late. I created configurations, but as I'm terrible with **nginx** tests using Perl, may I just send you the configurations? :sweat_smile:

HanadaLee:
The abbreviation for `response` is usually `rep` or `resp`, or consider not using the abbreviation and directly name the directive: `modsecurity_skip_request_body_filter` and `modsecurity_skip_response_body_filter`

thekief:
@HanadaLee thank you for looking over the changes. I use `resp` to abbreviate `response` now

airween:
@thekief could you rebase your PR? Then we can find some solution for testing.

thekief:
Done. I also had to add another commit to address the number of directives mentioned in the readme.

HanadaLee:
I found several blocking issues in the current implementation. In particular, the request-side directive currently disables much more than request-body inspection, and the new configuration fields do not inherit correctly.

1. modsecurity_skip_req_body_filter skips the entire ModSecurity transaction

In ngx_http_modsecurity_access_handler(), the new check returns before the ModSecurity context and transaction are created:
```
if (mcf->skip_req_body_filter == 1) {
    dd("Skipping request body filter");
    return NGX_DECLINED;
}
```
This does not only skip request-body buffering. It also skips:

connection processing;
URI processing;
request-header processing;
phase 1 rules;
response-header processing;
response-body processing;
transaction logging.

Because no context is created, the response filters and log handler subsequently treat ModSecurity as disabled for the entire request.

This contradicts the intended behavior described in issue #343, which is to skip body inspection while preserving the other ModSecurity checks.

The skip check should be moved after the transaction, URI, and request headers have been processed. It should only bypass ngx_http_read_client_request_body() and the body-appending logic.

The intended phase 2 behavior should also be defined explicitly. It may still be necessary to call:
```
msc_process_request_body(ctx->modsec_transaction);
```
without appending body data, so that phase 2 rules that do not depend on request-body variables can still run.

2. The new directives do not inherit from http or server

The fields are allocated through ngx_pcalloc(), so they initially have the value 0. They are not initialized to NGX_CONF_UNSET, and they are not handled in ngx_http_modsecurity_merge_conf().

As a result, a configuration such as:

server {
    modsecurity_skip_req_body_filter on;

    location /upload {
        proxy_pass http://backend;
    }
}

does not enable the option inside /upload, because the child location has its own zero-initialized value.

Please initialize both fields in ngx_http_modsecurity_create_conf():
```
conf->skip_req_body_filter = NGX_CONF_UNSET;
conf->skip_resp_body_filter = NGX_CONF_UNSET;
```
and merge them in ngx_http_modsecurity_merge_conf():
```
ngx_conf_merge_value(c->skip_req_body_filter,
                     p->skip_req_body_filter, 0);

ngx_conf_merge_value(c->skip_resp_body_filter,
                     p->skip_resp_body_filter, 0);
```
Without this, the documented http and server contexts do not work as expected.

3. The response body filter no longer restores a cleared module context

The PR changes:
```
ctx = ngx_http_modsecurity_get_module_ctx(r);
```
to:
```
ctx = ngx_http_get_module_ctx(r, ngx_http_modsecurity_module);
```
The custom ngx_http_modsecurity_get_module_ctx() helper is important because it restores the transaction context from the request-pool cleanup handlers when the module context has been cleared, for example after an internal redirect.

Using ngx_http_get_module_ctx() directly can therefore cause response-body inspection to be silently skipped after an internal redirect.

This is a regression even when both new directives are disabled. The original helper should continue to be used.

4. Response buffering is still requested when response inspection is skipped

The response header filter unconditionally sets:
```
r->filter_need_in_memory = 1;
```
The new skip_resp_body_filter check only happens later in the body filter. Therefore, Nginx is still told that the response body must be kept in memory even when response-body inspection is disabled.

To achieve the intended resource-saving behavior, the configuration should also be checked in the header filter:
```
if (!mcf->skip_resp_body_filter) {
    r->filter_need_in_memory = 1;
}
```
Response-header processing should still continue normally.

The phase 4 behavior should also be clarified. Completely skipping msc_process_response_body() skips all phase 4 rules, including rules that may not depend on RESPONSE_BODY.

5. Documentation and tests

There are also several documentation issues:

The README says the connector adds seven directives, but there are six existing directives plus two new ones, so the total should be eight.
The new directives are documented after the Contributing section instead of in the directive list.
The response directive says that it skips caching of the “request body”.
The SecRequestBodyAccess description should say that it is set to Off.

Tests should cover at least:

phase 1 rules still running when request-body inspection is skipped;
request-body rules not receiving body data;
response-header and logging phases still running;
inheritance and overriding at http, server, and location levels;
response-body inspection after an internal redirect;
streaming behavior with proxy_request_buffering off.

thekief:
Hey @HanadaLee ,
thank you for pointing that out. It looks like I messed up the rebase back in the day and never pushed the full fix again ...

Regarding the documentation and the number of directives: 
I have a second PR open, which I expected to me merged prior to this one but it also got stuck in a limbo.

tbh, I'm not involded with any `nginx` projects anymore so I cannot really test it either without re-building a test setup and I don't really have time for it either. 

If you want, I can check if I can find the intial patch and maybe you can get it merged. Otherwise, I'd close it as I don't want to keep it open without any progress.

HanadaLee:
> Hey @HanadaLee , thank you for pointing that out. It looks like I messed up the rebase back in the day and never pushed the full fix again ...
> 
> Regarding the documentation and the number of directives: I have a second PR open, which I expected to me merged prior to this one but it also got stuck in a limbo.
> 
> tbh, I'm not involded with any `nginx` projects anymore so I cannot really test it either without re-building a test setup and I don't really have time for it either.
> 
> If you want, I can check if I can find the intial patch and maybe you can get it merged. Otherwise, I'd close it as I don't want to keep it open without any progress.

Alright, if you don't mind, I'll probably submit a new PR based on your work, since I really need this feature.

airween:
@HanadaLee: first of all, thank you for your participation.

@thekief and @HanadaLee: I would like to let you know guys that I've been working on v4, where there will be some new features in API, eg. the application can ask the engine does it need to send the request body for the check (based on `SecResponseBodyMimeType` settings).

Also please note, that there is a known wrong behavior, described in [ModSecurity/#2465](https://github.com/owasp-modsecurity/ModSecurity/issues/2465), which needs to be fixed before we touch this part of API. I plan to fix it in v4 too.

HanadaLee:
> @HanadaLee: first of all, thank you for your participation.
> 
> @thekief and @HanadaLee: I would like to let you know guys that I've been working on v4, where there will be some new features in API, eg. the application can ask the engine does it need to send the request body for the check (based on `SecResponseBodyMimeType` settings).
> 
> Also please note, that there is a known wrong behavior, described in [ModSecurity/#2465](https://github.com/owasp-modsecurity/ModSecurity/issues/2465), which needs to be fixed before we touch this part of API. I plan to fix it in v4 too.

I am excited to see a brand-new version of ModSecurity on the horizon. I am currently planning to add support for the following features and hope they can be considered during development:
1. `modsecurity_bypass $xxxx` (Skip ModSecurity functionality based on certain Nginx variables.)
2. `modsecurity_dry_run on | off` (No interception. only status logging.)
3. `modsecurity_intervention_mode follow | deny | log` -> only log, do not action (Actually, I'm still considering it, because it seems very similar to a dry run.) 
5. `$modsecurity_status`: PASSED, REJECTED, REJECTED_DRY_RUN（like $limit_req_status）
6. `$modsecurity_intervention_status`: Returns the status code for ModSecurity intervention; recorded as `NGX_OK` (0) if no intervention occurred.   
7. `$modsecurity_triggered_rules`: https://github.com/owasp-modsecurity/ModSecurity-nginx/pull/374 (some bug, deferred)
8. Request body inspection bypass: https://github.com/owasp-modsecurity/ModSecurity-nginx/pull/356 (some bug, deferred)
9. H3 support: https://github.com/owasp-modsecurity/ModSecurity-nginx/pull/364 (some bug, deferred)
10. ​​Configuration inheritance: https://github.com/owasp-modsecurity/ModSecurity-nginx/pull/291 (some bug, deferred)

thekief:
@airween is this PR something you'd see as something that directly relates to/affects https://github.com/owasp-modsecurity/ModSecurity/issues/2465 ?

airween:
> I am excited to see a brand-new version of ModSecurity on the horizon.

Just for the record: it's not a brand new :smiley:.

But we definitely need to extend the API, therefore we need to increase the version. 

> is this PR something you'd see as something that directly relates to/affects https://github.com/owasp-modsecurity/ModSecurity/issues/2465 ?

No, that's different. We should solve that inside of the library.

I'll reply your listed items later.

thekief:
@HanadaLee I couldn't find the patch but I'll redevelop it. I will need until end of next week likely though, as I do not have an existing C development environment on my current PC.

tomsommer:
Following up on the two blocking points from the review (the request-side directive returning before the transaction is created, which disables the whole WAF for the location rather than just the body, and the new fields not being merged, so the setting does not inherit): I have opened #391 with the same goal for the request side, implemented so that the transaction is still created, `msc_process_request_body()` still runs (phase 2 rules on `ARGS` and headers keep working), the flag is merged with `ngx_conf_merge_value`, and with a Test::Nginx file covering the default, `off`, `off` with `proxy_request_buffering off`, inheritance and override. #395 does the response side and additionally stops setting `filter_need_in_memory`, which is what #206 asks for.

Both reference this PR as prior art. If you would rather fold that into #356 instead, the branches are on my fork and free to take. Prepared during a performance review of the connector done with Claude Fable 5.1.

airween:
To all: thanks for participated this PR.

@thekief, @HanadaLee: could you check @tomsommer's mentioned PR's?

@tomsommer: I started to review all pending PR's for Nginx connector (totally accidentally today...), and I saw you've sent so many new PR's. Thank you for them, hopefully sooner or later I'll finish to review all of them :smiley:.

PR #360, opened 2025-08-13
added disable error log support

fatihusta:
modified old PR

https://github.com/owasp-modsecurity/ModSecurity-nginx/pull/327

PR #361, opened 2025-09-22
Hook into Nginx NGX_HTTP_ACCESS_PHASE

### LINKED ISSUE #293: ModSecurity header phase runs before nginx rate limiting
As a user, we would like to skip the cost of ModSecurity rule processing for requests that are rate-limited by nginx. Today, the ModSecurity header phase is processed before the [ngx_http_limit_req_module](https://nginx.org/en/docs/http/ngx_http_limit_req_module.html) considers rate limiting.

This is due to how each module hooks into [the nginx request phases](https://nginx.org/en/docs/dev/development_guide.html#http_phases). The ngx_http_limit_req_module enforces rate limiting [in the NGX_HTTP_PREACCESS_PHASE](https://github.com/nginx/nginx/blob/765ef1098d169622ada67eb4eccad1d45d47a61c/src/http/modules/ngx_http_limit_req_module.c#L1094-L1099), but ModSecurity-nginx processes the ModSecurity header phase earlier, [during the NGX_HTTP_REWRITE_PHASE](https://github.com/SpiderLabs/ModSecurity-nginx/blob/d59e4ad121df702751940fd66bcc0b3ecb51a079/src/ngx_http_modsecurity_module.c#L549-L555).

Some workarounds are:
- Move all rules to the ModSecurity request body, but this is not a great option when using large, shared rulesets like the [OWASP ModSecurity Core Rule Set](https://owasp.org/www-project-modsecurity-core-rule-set/)
- Make a custom build that runs the ModSecurity request header phase in the PREACCESS phase. This is the same nginx phase used to run the ModSecurity request body phase, but unlike the request body phase, the header phase wouldn't have to wait to receive the request body.

Even with these workarounds, it seems like it might make sense to be able to rate-limit prior to ModSec processing. Is this something we could consider changing or making configurable?

tomsommer:
Hook into Nginx `NGX_HTTP_ACCESS_PHASE` instead of `NGX_HTTP_PREACCESS_PHASE` and `NGX_HTTP_REWRITE_PHASE`.

Having ModSecurity run before `NGX_HTTP_ACCESS_PHASE` cripples Nginx during DDOS attacks, because ModSecurity runs before Nginx limit_* functions and consumes too much CPU/memory even in phase:1.

Fixes #293


airween:
Hi @tomsommer!

Thanks for this PR. Could you add some test(s) (see others [here](https://github.com/owasp-modsecurity/ModSecurity-nginx/tree/master/tests)) to check that the behavior is that the modification expects? I mean set up the engine with some very low rate limit and try to send more parallel requests than that. Or you can add a new job in our [workflow](https://github.com/owasp-modsecurity/ModSecurity-nginx/tree/master/.github/workflows).

Thanks again!

tomsommer:
What do you want to test? It sounds like you want to test if Nginx limit_* works? 
The current tests should be able to test if this PR breaks anything.

airween:
> What do you want to test? It sounds like you want to test if Nginx limit_* works?

No, I want to test if this patch solves the mentioned issue and that you wrote (in a removed comment): _"nginx cannot survive a DDOS if modsec is before limit*"_. It would be nice to see that if the user sets a limit and requests exceed then Nginx drops the connections and the requests do not reach ModSecurity.

> The current tests should be able to test if this PR breaks anything.

Yes, that's a necessary but not sufficient condition.

tomsommer:
I did a lot of testing and I had to merge the rewrite stuff into the access handler, instead of registering two handlers in the same phase (I don't know if it's the order, or whatever), but it works now. It runs after limit_*

airween:
Hi @tomsommer,


first of all, many thanks for your fix, and apologize for the late response.

I think this is a very important solution.

> I did a lot of testing and I had to merge the rewrite stuff into the access handler, instead of registering two handlers in the same phase (I don't know if it's the order, or whatever), but it works now. It runs after limit_*

I saw the tests - thanks for them - which justify the code works as well. Thank you for them too.

I'm still a little worried (I'm not an Nginx developer), because:
* I can't see (and probably nobody) what other side effects will be there (eg the rewrite hooks; I've made some researches and found that rewrite happens also in pre-access; I don't know how important is that from this aspect)
* I'm sure the original authors (including @defanator, who was an Nginx developer) knew what they were doing, didn't put the connector in pre-access by accident, but on purpose.

If we are lucky now, @defanator will see this comment and may be he will explain the reason.

If there won't be any response or other discussion, I'll merge this into the master, and hopefully many people will test that, which will provide some feedback (as it happened in case of #334 which was reverted in #344).

airween:
Unfortunately, no one has responded, and I don't see the point in waiting any longer. I'm merging this PR now, and hope a few users start to use this.

Thanks again, @tomsommer!

tomsommer:
I'll test this and #360

PR #364, opened 2025-12-04
Add HTTP/3 Host header support for ModSecurity

### LINKED ISSUE #305: 'Host' header missing when HTTP3 is in use prevents "SecRule REQUEST_HEADERS:Host" rules from being effective.
Hello Team and thanks for your work on the modsec connector for Nginx.
We built nginx with modsec and the modsec connector using the official 'nginx-quic' branch for implementing HTTP3. We use nginx primarily as a global reverse proxy server for our web infrastructure, providing SSL/TLS encryption for our web apps as well as threats protection thanks to modsec.
Everything is working fine apart from several rules we implemented which at the time seemed ineffective, for some unknown reasons.
After some debugging (by enabling modsec's audit logs), it appears that the 'Host' header is missing when the clients request via HTTP3 (that header is not present in modsec's audit logs). That would explain our issue: since that header is unavailable to modsec, all rules like "SecRule REQUEST_HEADERS:Host" are ineffective when clients connect via HTTP3. We then checked with both HTTP1.1/2: the Host header is present in the audit logs, and our rules are applied successfully as they should. We tried to find a workaround by using the 'headers-more' module from openresty and applying custom inbound headers. However it seems like modsec's processing is done prior to our custom headers being injected, which makes such workaround hopeless. Not quite sure what to do next. It's also possible that the Host header not being present is an HTTP3 specification, possibly. It is to note that rules using REQUEST_URI are working normally via all protocols, including HTTP3. Any possible to suggest? Is anybody else experiencing the same? Thanks in advance for your help!

AMoo-Miki:
ModSecurity cannot see the Host header in HTTP/3 requests because HTTP/3 uses the `:authority` pseudo-header, which nginx parses into `r->headers_in.server` but doesn't add to the headers list.

This commit:
- Adds `NGX_HTTP_VERSION_30` case to `http_version` switch
- Manually extracts Host from `r->headers_in.server` for HTTP/3 requests
- Adds Host header to ModSecurity transaction before processing other headers

Fixes #305 false positives from OWASP CRS rule 920280 (Missing Host Header) on HTTP/3 connections.

Tested with nginx 1.29.3 and ModSecurity 3.0.13.

### DIFF
--- a/src/ngx_http_modsecurity_access.c
+++ b/src/ngx_http_modsecurity_access.c
@@ -199,6 +199,11 @@ ngx_http_modsecurity_access_handler(ngx_http_request_t *r)
                 http_version = "2.0";
                 break;
 #endif
+#if defined(nginx_version) && nginx_version >= 1025000
+            case NGX_HTTP_VERSION_30 :
+                http_version = "3.0";
+                break;
+#endif
             default :
                 http_version = ngx_str_to_char(r->http_protocol, r->pool);
                 if (http_version == (char*)-1) {
@@ -233,9 +238,26 @@ ngx_http_modsecurity_access_handler(ngx_http_request_t *r)
         }
 
         /**
-         * Since incoming request headers are already in place, lets send it to ModSecurity
+         * HTTP/3 uses :authority pseudo-header instead of Host header and nginx
+         * parses it into r->headers_in.server (see ngx_http_v3_request.c#L982)
+         * but doesn't add it to the headers list, so ModSecurity never sees it.
          *
+         * Per RFC 9114 §4.3.1, when an HTTP/3 request is normalized to HTTP/1.1,
+         * a `Host` header must be generated from the `:authority` pseudo-header
+         * if `Host` is absent. This code does not check for a pre-existing `Host`
+         * header because, if present, it will be overwritten by the subsequent
+         * call to `msc_add_n_request_header()`.
          */
+        if (strcmp(http_version, "3.0") == 0 && r->headers_in.server.len > 0) {
+            dd("adding Host header from :authority: %.*s",
+               (int)r->headers_in.server.len, r->headers_in.server.data);
+
+            msc_add_n_request_header(ctx->modsec_transaction,
+                (const unsigned char *)"Host", 4,
+                (const unsigned char *)r->headers_in.server.data,
+                r->headers_in.server.len);
+        }
+
         ngx_list_part_t *part = &r->headers_in.headers.part;
         ngx_table_elt_t *data = part->elts;
         ngx_uint_t i = 0;
--- a/tests/modsecurity-h3.t
+++ b/tests/modsecurity-h3.t
@@ -1,0 +1,143 @@
+#!/usr/bin/perl
+
+# Tests for ModSecurity module (HTTP/3).
+# Tests that Host header from :authority pseudo-header is passed to ModSecurity.
+
+###############################################################################
+
+use warnings;
+use strict;
+
+use Test::More;
+
+BEGIN { use FindBin; chdir($FindBin::Bin); }
+
+use lib 'lib';
+use Test::Nginx;
+use Test::Nginx::HTTP3;
+
+###############################################################################
+
+select STDERR; $| = 1;
+select STDOUT; $| = 1;
+
+my $t = Test::Nginx->new()->has(qw/http http_v3/)
+	->has_daemon('openssl');
+
+$t->write_file_expand('nginx.conf', <<'EOF');
+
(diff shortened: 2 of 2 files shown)


airween:
Hi @AMoo-Miki,

first of all, thank you for your PR.

I am sure that this is a very important step towards being able to handle HTTP/3 requests. But I'm not sure this is the right way.

Please do not misunderstand me, I don't want to reject this PR, but I think we need to discuss how can we handle HTTP/3 headers correctly.

I think adding a header with the name `Host` so that it is not the host header is not accurate and might be misleading. I mean in case of HTTP/3, the header `authority` should be added as `authority` - but I'm not sure that, this is why I think we should discuss that.

Also, please add some tests to this new feature.

AMoo-Miki:
@airween thanks for looking into this.

I totally get where you are coming from but if I put my "salesman hat" on, I could say adding it as a "Host" header here isn't really fabricating data; it's exposing the same value to ModSecurity's `REQUEST_HEADERS` collection so that rules checking `REQUEST_HEADERS:Host` (like CRS 920280) work correctly.

Here are some other options I considered:
1. Create a fake header called "Authority" in the connector because NGINX hands `:authority` in `r->headers_in.server` and not headers. This is a bad idea because it is not a standard key and can conflict or be confused with actual headers in the request. Additionally, all rules that check for `Host` will need to be updated to check for `Authority` too.
2. Create `msc_set_request_authority` in ModSecurity which would populate `REQUEST_HEADERS:Host`. This would add complications because the lib would have to deal with the possibility of conflicting Host headers.
3.  Create `msc_set_request_authority` in ModSecurity which would populate `REQUEST_AUTHORITY` but then every existing rule that checks for `REQUEST_HEADERS:Host` would need to be updated to also check  `REQUEST_AUTHORITY`.
4. Similar to 3 but on a larger scale: introduce the concept of pseudo-headers in ModSecurity; this would the future-safe thing to do but is a substantial undertaking. However, this wouldn't help this connector as NGINX translates other pseudo-headers to request props already.

What really helped me choose was [RFC 9114, Section 4.3.1](https://www.rfc-editor.org/rfc/rfc9114.html#section-4.3.1-2.9): 

> An intermediary that converts an HTTP/3 request to HTTP/1.1 MUST create a `Host` field if one is not present in a request by copying the value of the `:authority` pseudo-header field.

I interpreted that as it being necessary for this connector to populate `Host` with `:authority` when it is missing. You might notice that my change doesn't "waste" time checking if `Host` header is missing; this is because if it somehow does exist, the next block will overwrite it when calling `msc_add_n_request_header()` for the actual headers -- Lemme know if you object.

This approach ensures consistent behavior across HTTP/1.1, HTTP/2, and HTTP/3 without requiring rule modifications.

~PS, working on adding tests.~
Added tests.

HanadaLee:
1. `msc_add_n_request_header` appears that headers may be added redundantly. If an HTTP/3 request carries both `:authority` and `Host` headers, it might be logged as having two `Host` headers, ultimately causing certain rules to trigger false positives (since duplicate `Host` headers are disallowed).
2. `strcmp(http_version, "3.0") == 0` can be directly replaced with `r->http_version == NGX_HTTP_VERSION_30` to reduce the overhead of string matching.
3. The code written into the H3 header should also use macro `#if defined(nginx_version) && nginx_version >= 1025000`.

airween:
@AMoo-Miki,

please consider to review @HanadaLee's comments above, especially the 1st point:
> 1. `msc_add_n_request_header` appears that headers may be added redundantly. If an HTTP/3 request carries both `:authority` and `Host` headers, it might be logged as having two `Host` headers, ultimately causing certain rules to trigger false positives (since duplicate `Host` headers are disallowed).

Beside of that, your [comment](https://github.com/owasp-modsecurity/ModSecurity-nginx/pull/364/changes#diff-e6bc2e7283bc6b7b9add2e18d5ea92fcdaf3f235083e257a7ac208f9d24f1b71R247-R248) probably is not true:

>           * if `Host` is absent. This code does not check for a pre-existing `Host`
>           * header because, if present, it will be overwritten by the subsequent

There is no overwritten.

And one more thing: you use `r->headers_in.server` [here](https://github.com/owasp-modsecurity/ModSecurity-nginx/pull/364/changes#diff-e6bc2e7283bc6b7b9add2e18d5ea92fcdaf3f235083e257a7ac208f9d24f1b71R257), which is a truncated value - I mean if the `Host` (`:authority`) contains a port (eg `example.com:8080`) - then the value will be `example.com`. Which is not the consistent behavior across HTTP/1.1, 2 and 3 as you mentioned.

Thank you!

PR #366, opened 2025-12-23
Remove not needed code fragment

tomsommer:
@airween I'm running `master` in production on 400+ servers, both #360 and #361 works fine. 

AI claims there are some odd things in the code that could be optimized, but that's beyond my scope :)


airween:
Thank you @tomsommer, merging now.

PR #367, opened 2026-02-04
Do not allocate new string for common HTTP methods

### LINKED ISSUE #123: Per-location rule exclusion
modsecurity on;
modsecurity_rules '
SecRuleRemoveById XXXXXX
';

This is kind of inconvenient.

Regards,
D

JakubOnderka:

### DIFF
--- a/src/ngx_http_modsecurity_access.c
+++ b/src/ngx_http_modsecurity_access.c
@@ -212,15 +212,41 @@ ngx_http_modsecurity_access_handler(ngx_http_request_t *r)
                 break;
         }
 
+        // Do not allocate new string for common methods
+        const char *n_method;
+        switch (r->method) {
+            case NGX_HTTP_GET:
+                n_method = "GET";
+                break;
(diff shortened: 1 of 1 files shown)


airween:
Hi @JakubOnderka,

airween:
Hi @JakubOnderka,

beside the note above, I found some other things:
* `NGX_HTTP_CONNECT` was introduced in Nginx 1.21.1. A few older systems contain 1.18, this change breaks the build for that; please consider to use a guard macro ([lines](https://github.com/owasp-modsecurity/ModSecurity-nginx/pull/367/changes#diff-e6bc2e7283bc6b7b9add2e18d5ea92fcdaf3f235083e257a7ac208f9d24f1b71R230-R232) in the patch
* Nginx refuses the `CONNECT` method before the access phase, so the mentioned code above never will be executed
* there are some other, more frequent methods: `PUT`, `DELETE`, `PATCH` - which are completely missing

Do you want to fix these?

PR #368, opened 2026-02-04
Avoid allocation for server address

### LINKED ISSUE #123: Per-location rule exclusion

JakubOnderka:

### DIFF
--- a/src/ngx_http_modsecurity_access.c
+++ b/src/ngx_http_modsecurity_access.c
@@ -147,26 +147,20 @@ ngx_http_modsecurity_access_handler(ngx_http_request_t *r)
 #endif
 
         ngx_str_t s;
-        u_char addr[NGX_SOCKADDR_STRLEN];
-        s.len = NGX_SOCKADDR_STRLEN;
-        s.data = addr;
+        u_char server_addr[NGX_SOCKADDR_STRLEN + 1];
+        s.len = 0;
+        s.data = server_addr;
         if (ngx_connection_local_sockaddr(r->connection, &s, 0) != NGX_OK) {
             return NGX_HTTP_INTERNAL_SERVER_ERROR;
         }
-
-        const char *server_addr = ngx_str_to_char(s, r->pool);
-        if (server_addr == (char*)-1) {
-            return NGX_HTTP_INTERNAL_SERVER_ERROR;
-        }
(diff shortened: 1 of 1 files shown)


airween:
Hi @JakubOnderka,

src/ngx_http_modsecurity_access.c
```diff
@@ -147,26 +147,20 @@ ngx_http_modsecurity_access_handler(ngx_http_request_t *r)
 #endif
 
         ngx_str_t s;
-        u_char addr[NGX_SOCKADDR_STRLEN];
-        s.len = NGX_SOCKADDR_STRLEN;
-        s.data = addr;
+        u_char server_addr[NGX_SOCKADDR_STRLEN + 1];
+        s.len = 0;
```
airween, line 151:
```suggestion
        s.len = NGX_SOCKADDR_STRLEN; /* in: buffer capacity, out: written length */
```
please consider to set this value

src/ngx_http_modsecurity_access.c
```diff
@@ -147,26 +147,20 @@ ngx_http_modsecurity_access_handler(ngx_http_request_t *r)
...
+        s.data = server_addr;
         if (ngx_connection_local_sockaddr(r->connection, &s, 0) != NGX_OK) {
             return NGX_HTTP_INTERNAL_SERVER_ERROR;
         }
-
-        const char *server_addr = ngx_str_to_char(s, r->pool);
-        if (server_addr == (char*)-1) {
-            return NGX_HTTP_INTERNAL_SERVER_ERROR;
-        }
+        server_addr[s.len] = '\0'; // ngx_str_t is not null terminated, so we have insert \0 manually
```
airween, line 156: What if the address is a unix socket?
Please consider to check this suggestion:
```suggestion
        if (s.len > NGX_SOCKADDR_STRLEN) {
            return NGX_HTTP_INTERNAL_SERVER_ERROR;
        }
        server_addr[s.len] = '\0'; // ngx_str_t is not null terminated, so we have to insert \0 manually
```

airween:
@JakubOnderka,

beside my first comment, please check my notes above and **please add tests**.

Thank you!

PR #370, opened 2026-02-04
Do not check static string length

### LINKED ISSUE #123: Per-location rule exclusion

JakubOnderka:


airween:
Hi @JakubOnderka,

Also please take a look at the QA failures.

airween:
@JakubOnderka,

please take a look at my previous [comment](https://github.com/owasp-modsecurity/ModSecurity-nginx/pull/370#issuecomment-3862474407) and Copilot's review.

Thank you!

PR #198, opened 2020-05-01
Support audit log rotation on SIGHUP and SIGUSR1

brandonpayton:
This is a PR that uses SpiderLabs/ModSecurity#2304 to support audit log rotation when nginx reloads config or reopens log files.

Thanks to @defanator for [providing a proof-of-concept](https://github.com/SpiderLabs/ModSecurity-nginx/issues/121#issuecomment-442416602)!

I tested this by:
1. Using `lsof` to observe nginx has the audit log open
1. Moving the audit log file
1. Using `lsof` to confirm the nginx now references the moved file
1. Signaling the nginx master process with SIGHUP or SIGUSR1.
1. Testing the same things with `nginx -s reload` and `nginx -s reopen`

When I wasn't convinced of the thread-safety of audit log writes and log reopen within nginx, I also attempted to test it with the following script using GNU parallel:
`parallel -j+0 ./test-reopen.sh ::: {1..10000}`

```bash
#!/bin/bash

set -euo pipefail

read s < /dev/urandom
TEST=$(( ${#s} % 11 ))

if [[ TEST -eq 0 ]]; then
	echo "$1 reopen with SIGUSR1"
	kill -USR1 "$(pgrep -P1 nginx)"
elif [[ TEST -eq 1 ]]; then
	echo "$1 reload with SIGHUP"
	kill -HUP "$(pgrep -P1 nginx)"
else
	echo "$1 triggering modsec logging"
	# Matches a silly rule created for testing
	curl --silent http://localhost/index.html?asdf 2>&1 >/dev/null &
fi

pgrep -d", " nginx
```

This test didn't hurt, but due to nginx's primarily single-threaded architecture, I believe log reopen is thread-safe here by default.

resolves #121

PR #372, opened 2026-03-24
Support log reopen on SIGHUP and SIGUSR1

### LINKED ISSUE #121: reopen audit log on SIGUSR1 and SIGHUP
Modsecurity should reopen audit log on these two signals for proper logrotate operation.

AntonDemushkin:
This PR uses new ModSecurity API implemented in owasp-modsecurity/ModSecurity#3521

The CI build fails because the ModSecurity API isn't released yet :point_up: 

There is another PR with a similar code -- #198 -- but its code is outdated and uses a different API. In any case this PR is based on work in #198, so thank you @brandonpayton for the initial work on that.

I've tested the code using testing notes from #198. I hope we'll see these changes released soon!

## Side notes for reviewers

`msc_rules_reopen_logs()` only exists in rules_set.h (the newer API) — it's not available in the older rules.h at all. So the ``#ifdef MSC_USE_RULES_SET` guard is necessary. Without it the build would fail against older
  libmodsecurity versions that don't have this function.

## Links

This PR depends on code from owasp-modsecurity/ModSecurity#3521
Resolves #121
Similar PR #198

### DIFF
--- a/src/ngx_http_modsecurity_common.h
+++ b/src/ngx_http_modsecurity_common.h
@@ -109,6 +109,7 @@ typedef struct {
     ngx_uint_t                 rules_inline;
     ngx_uint_t                 rules_file;
     ngx_uint_t                 rules_remote;
+    ngx_open_file_t           *log_reopen;
 } ngx_http_modsecurity_main_conf_t;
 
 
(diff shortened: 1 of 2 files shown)


airween:
@AntonDemushkin,

could you take a look at Copilot's review?

Current status: we wait to merge [ModSecurity/3582](https://github.com/owasp-modsecurity/ModSecurity/pull/3582) then release a new version of library.

PR #374, opened 2026-04-22
Add modsecurity log vars

meirdev:
## what

Add two variables to the module: `$modsecurity_intervention` and `$modsecurity_triggered_rules`.

* `$modsecurity_intervention`: set to 1 if ModSecurity triggered a disruptive intervention.
* `$modsecurity_triggered_rules`: a comma-separated list of matched rule IDs.

## why

NGINX logs already contain a lot of useful data, but they don't include ModSecurity events. That makes it hard to correlate what you see in NGINX logs with what shows up in ModSecurity audit logs.

If you're sending logs to an "append-only" databsae, theres no easy way to correlate the two later. And trying to sync them beforehand usually means building a complex buffering and matching system.

### DIFF
--- a/src/ngx_http_modsecurity_module.c
+++ b/src/ngx_http_modsecurity_module.c
@@ -46,10 +46,12 @@ static void ngx_http_modsecurity_cleanup_rules(void *data);
 static ngx_http_variable_t ngx_http_modsecurity_vars[] = {
 
     { ngx_string("modsecurity_intervention"), NULL,
-      ngx_http_modsecurity_intervention_variable, 0, 0, 0 },
+      ngx_http_modsecurity_intervention_variable,
+      0, NGX_HTTP_VAR_NOCACHEABLE, 0 },
 
     { ngx_string("modsecurity_triggered_rules"), NULL,
-      ngx_http_modsecurity_triggered_rules_variable, 0, 0, 0 },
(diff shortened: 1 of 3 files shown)


airween:
@meirdev: could you merge your branch with the current master? @Easton97-Jens fixed the workflow test, probably your PR will be okay now.

src/ngx_http_modsecurity_module.c
```diff
@@ -564,6 +585,112 @@ ngx_module_t ngx_http_modsecurity_module = {
...
+    int64_t *ids;
+    u_char  *buf, *p, *end;
+
+    ctx = ngx_http_modsecurity_get_module_ctx(r);
+    if (ctx == NULL || ctx->modsec_transaction == NULL) {
+        v->not_found = 1;
+        return NGX_OK;
+    }
+
+    size = msc_get_rules_messages_size(ctx->modsec_transaction);
```
airween, line 626: Here (and below when you fill variable `written`) assume `msc_get_rules_messages_...` function exists.

These functions appeared in the library in 3.0.15, with [this](https://github.com/owasp-modsecurity/ModSecurity/commit/ddf6234c49d64507e1ef0865755c147c108375d4) commit, so it would be nice to add a guard to avoid a build error with older library version.

airween (changes requested):

PR #377, opened 2026-05-11
Implement Phase 4 handling for ModSecurity-nginx

Easton97-Jens:
### What changed

This PR introduces and stabilizes dedicated `phase:4` handling for the nginx ModSecurity connector.

Main additions include:

- Support for configurable phase4 handling modes:
  - `minimal`
  - `safe`
  - `strict`
- Explicit late-intervention handling when response headers were already sent
- Dedicated phase4 JSON logging via:
  - `modsecurity_phase4_log`
- Configurable content-type scoping through:
  - `modsecurity_phase4_content_types_file`
- Additional parser, logging, HTTP/2, proxy, and regression test coverage
- Production-oriented example configurations and documentation

Additional fixes included in this PR:

- Fixed multiple phase4-related build issues
- Corrected raw-rule and status logging order
- Added missing parser and regression test cases
- Improved validation and configuration parsing behavior

---

### Why these changes

`phase:4` operates on response bodies. At this stage, nginx may already have sent response headers or even part of the response body.

Because of this technical limitation, actions such as:

- `deny`
- `status`
- `redirect`

cannot always be converted into a clean client-visible HTTP response once headers are already sent.

This PR introduces explicit and predictable handling for those late-intervention scenarios:

- `minimal`
  → downgrade to `log_only`
- `safe`
  → downgrade to `log_only` (default-safe production behavior)
- `strict`
  → optional `connection_abort`

The goal is to:

- avoid false guarantees around late blocking behavior,
- make phase4 handling transparent and deterministic,
- improve operational visibility,
- provide safer production defaults,
- clearly separate production behavior from test/demo behavior.

---

### Logging and observability

The new phase4 logging implementation provides structured JSON-lines logging including fields such as:

- intervention mode
- requested vs actual action
- header-sent state
- response status
- content type
- rule ID
- downgrade reason

Tests also verify that response-body payload data is not leaked into phase4 logs.

---

### Testing

This PR adds and extends coverage for:

- parser validation
- regression handling
- proxy paths
- HTTP/2 handling
- late-intervention scenarios
- content-type scoping
- logging behavior

---

### Documentation

Comprehensive documentation for the implemented behavior was added, including:

- operational limitations of `phase:4`
- behavior differences between modes
- production configuration examples
- logging semantics
- known technical boundaries and guarantees

### DIFF
--- a/.github/scripts/resolve-nginx-ref.sh
+++ b/.github/scripts/resolve-nginx-ref.sh
@@ -1,0 +1,34 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+VERSION="${NGINX_VERSION:-}"
+
+if [[ -z "${VERSION}" ]]; then
+  echo "NGINX_VERSION is not set" >&2
+  exit 1
+fi
+
+echo "Requested NGINX_VERSION=${VERSION}"
+
+if [[ "${VERSION}" =~ ^[0-9]+\.[0-9]+\.x$ ]]; then
+  PREFIX="release-${VERSION%.x}."
+  REF=$(git ls-remote --tags https://github.com/nginx/nginx.git "refs/tags/${PREFIX}*" \
+    | sed -E 's#.*refs/tags/##; s#\^\{\}##' \
+    | grep -E "^${PREFIX}[0-9]+$" \
+    | sort -V \
+    | tail -n 1)
+elif [[ "${VERSION}" =~ ^release-[0-9]+\.[0-9]+\.[0-9]+$ || "${VERSION}" == "master" || "${VERSION}" == "main" ]]; then
+  REF="${VERSION}"
+else
+  echo "Unsupported NGINX_VERSION=${VERSION}" >&2
+  echo "Use e.g. 1.30.x, 1.29.x, 1.28.x or exact tag release-1.30.0" >&2
+  exit 1
+fi
+
+if [[ -z "${REF}" ]]; then
+  echo "No matching Nginx tag found for ${VERSION}" >&2
+  exit 1
+fi
+
+echo "Resolved Nginx ref: ${REF}"
+echo "ref=${REF}" >> "${GITHUB_OUTPUT}"
--- a/.github/workflows/test_new.yml
+++ b/.github/workflows/test_new.yml
@@ -1,33 +1,38 @@
-name: Quality Assurance
+name: Quality Assurance new
 
 on:
-  push:
   pull_request:
+  push: 
+  workflow_dispatch:
 
-  # Default: none
+
+# Default: none
 permissions: {}
 
 jobs:
   build-linux:
+    name: linux-${{ matrix.compiler }}-nginx-${{ matrix.nginx_version }}
     permissions:
       contents: read
     runs-on: ${{ matrix.os }}
     strategy:
+      fail-fast: false
       matrix:
-        os: [ubuntu-22.04]
+        os: [ubuntu-24.04]
         compiler: [gcc, clang]
+        nginx_version: ["1.30.x", "1.29.x", "1.28.x"]
     env:
+        NGINX_VERSION: ${{ matrix.nginx_version }}
         CC: "/usr/bin/${{ matrix.compiler }}"
         CXX: "/usr/bin/${{ matrix.compiler == 'gcc' && 'g' || 'clang' }}++"
         COMPDEPS: "${{ matrix.compiler == 'gcc' && 'gcc g++' || 'clang' }}"
     steps:
       - name: Setup Dependencies
         run: |
-          sudo dpkg --add-architecture i386
           sudo apt-get update -y -qq
-          sudo apt-get install -y make autoconf automake make libyajl-dev libxml2-dev libmaxminddb-dev libcurl4-gnutls-dev $COMPDEPS
+          sudo apt-get install -y make autoconf automake pcre2-utils libpcre2-dev libyajl-dev libxml2-dev libmaxminddb-dev libcurl4-gnutls-dev $COMPDEPS
       - name: Get libModSecurity source
-        uses: actions/checkout@v4
+        uses: actions/checkout@v6
         with:
           repository: owasp-modsecurity/ModSecurity
           path: ModSecurity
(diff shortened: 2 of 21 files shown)


airween:
Excellent PR, thanks @Easton97-Jens!

Please take a look at Copilot's suggestions.

I think it would be great to add some minimal description about the new directives to main README.

Other than that, I would like to test this PR locally, and play with it for a wile - I need some time for that.

Thanks again.

Easton97-Jens:
Thanks!
I’ve already addressed most of the feedback and also updated the README with a short description of the new directives.

The only remaining thing is one Copilot suggestion that I currently can’t apply cleanly without changing behavior, so I left it as-is for now.

Take your time testing the PR locally, and let me know if you find anything or have more feedback 🙂

Easton97-Jens:
https://github.com/Easton97-Jens/ModSecurity-test-Framework

I split out a small evidence/test framework that runs the same YAML-derived cases through real Apache and NGINX connector paths. It currently verifies ARGS, ARGS_NAMES, REQUEST_BODY, REQUEST_COOKIES, REQUEST_HEADERS, FILES, XML, AUDIT_LOG, RESPONSE_HEADERS, etc. RESPONSE_BODY remains xfail/mapped-only, and RAW argument collections are mapped-only with the current local v3 source because the ARGS_RAW family is not present there yet.

airween:
Thanks for the updates, @Easton97-Jens - finally, I was able to review this one (too).

The README additions and the Copilot round look good, and the test coverage here is genuinely thorough. I went through the branch locally, and there are two design questions I'd like to settle before we look at the implementation details, because they may change the shape of the patch.

## 1. The new behavior can't be turned off, and it isn't opt-in

`modsecurity_phase4_mode` accepts only `minimal|safe|strict`, there is no `off`, and the merge default is `safe`. The code path is unconditional: `ngx_http_modsecurity_body_filter()` always routes `ret < 0` into `ngx_http_modsecurity_phase4_handle_intervention()`, and `phase4_content_types` is never `NULL` after the merge because the built-in six-entry list is injected when no file is configured. So a user who never writes a single `modsecurity_phase4_*` directive still gets the new semantics after upgrading.

None of the three modes restores the previous behavior either. Before this PR, the `last_buf` path returned `ngx_http_filter_finalize_request(..., NGX_HTTP_INTERNAL_SERVER_ERROR)` and the mid-chain `ret < 0` was simply ignored; `strict` instead sets `r->connection->error = 1` and returns `NGX_ERROR` in both places. Close, but not the same - and not the default.

I'd like an `off` value that keeps the current behavior, and I'd prefer it to be the default, so the new handling is something you explicitly opt into. A silent change to what `phase:4 deny` does after upgrading a connector is not something we can ship without a flag.

## 2. `modsecurity_phase4_content_types_file` duplicates `SecResponseBodyMimeType`

This is the bigger one. MIME scoping for response bodies is already owned by the library via [`[SecResponseBodyMimeType](https://github.com/owasp-modsecurity/ModSecurity/wiki/Reference-Manual-(v3.x)#user-content-SecResponseBodyMimeType)`](https://github.com/owasp-modsecurity/ModSecurity/wiki/Reference-Manual-(v3.x)#user-content-SecResponseBodyMimeType) (plus `SecResponseBodyMimeTypesClear`). Having a second list in the connector means the same thing is configured in two places, with different syntax and different semantics.

And the connector's list isn't an alternative - it's a second, narrowing gate applied *after* the library has already decided:

1. `SecResponseBodyMimeType` decides whether the response body is buffered and inspected at all. If the type isn't listed, `RESPONSE_BODY` is never populated, the `phase:4` rule doesn't match, and no intervention is produced.
2. The connector's list is only consulted once an intervention already exists - i.e. once the library's MIME filter has already passed it.

So the connector list can only ever narrow, never widen. Concretely:

- The library default is `text/plain text/html text/xml`. The connector's built-in list also contains `application/json`, `application/xml` and `application/xhtml+xml` - dead entries in a default setup, because the library never gets that far. The module advertises a scope it cannot deliver.
- The reverse is worse. If someone deliberately sets `SecResponseBodyMimeType text/csv`, the rule fires and an intervention is produced, but the connector's default list has no `text/csv`, so even in `strict` mode it is silently downgraded to `log_only` with `reason:"content_type_not_in_scope"`. Enforcement disappears, and the only trace is a JSON log line that is off by default.
- The two mechanisms don't even agree on syntax: `SecResponseBodyMimeType` supports `*` and has `SecResponseBodyMimeTypesClear`; `ngx_http_modsecurity_phase4_validate_content_type()` explicitly rejects wildcards, and the connector only accepts a file, not an inline list.
- The connector reads `r->headers_out.content_type`, while the library used whatever it saw on its side - in proxy/filter chains those can differ.

My preference is to drop `modsecurity_phase4_content_types_file` from this PR and let `SecResponseBodyMimeType` remain the single place where MIME scoping is configured. If some connector-side scoping really is needed, it should derive from the library's list rather than from a parallel, hand-maintained file. Dropping it also removes the `content_type_missing` / `content_type_not_in_scope` branches, which cover cases the library wouldn't have inspected anyway.

I know I promised that the next library version (v4) will be able to provide a function which gives information about `SecResponseBodyMimeType`'s setting, so the connector can use that - if we arrived there...

--

The rest is implementation-level, and worth fixing regardless of how the two points above land.

## 3. Config inheritance is broken for `modsecurity_phase4_log` (and the content-type list)

`ngx_http_modsecurity_create_conf()` initializes these to `NULL`:

```c
conf->phase4_content_types = NULL;
conf->phase4_log_file = NULL;
```

but `ngx_http_modsecurity_merge_conf()` merges them with `ngx_conf_merge_ptr_value()`, which only copies from the parent when the child equals `NGX_CONF_UNSET_PTR`. Since they are `NULL`, the merge is a no-op and nothing is inherited from `http`/`server` into a `location`.

In practice: `modsecurity_phase4_log` set at `server` level produces no logging in any location. That's exactly what the shipped examples do - in `docs/examples/phase4-safe.conf` the log is declared in the `server` block while `location /` only has `proxy_pass`, so that example logs nothing. The tests don't catch it because every phase-4 directive sits inside the `location` block.

Please initialize both to `NGX_CONF_UNSET_PTR` (or merge them explicitly), and add a test that sets the log at `http`/`server` level and asserts it's used in a nested location. `phase4_mode` is fine - it uses `ngx_conf_merge_uint_value()` with `NGX_CONF_UNSET_UINT`.

## 4. `minimal` and `safe` are functionally identical

Both end in the same `log_only` path; only the `mode`/`reason` strings in the JSON log differ. Three documented modes with two distinct behaviors will confuse people. Either drop one, or give `minimal` a behavior that actually differs and document it.

## 5. Smaller points

- **CHANGES** isn't updated. Whatever we decide on point 1, this needs an entry.
- **README** now says "It adds eight directives" but lists nine. (The `four` in master was already wrong, so this is a good chance to just fix the count.)
- **`intervention` field in the JSON log**: `ngx_http_modsecurity_sanitize_intervention()` always yields either `redacted` or `id:present msg:present op:-`. Redaction is right, but the field name promises more than it delivers - consider renaming or dropping it now that `rule_id` exists.
- **`<ctype.h>`**: `isalnum()` / `isspace()` are locale-dependent; the rest of the module uses the `ngx_*` helpers. Please switch for consistency.
- **`ngx_http_modsecurity_phase4_log_event()`** dereferences `mcf` without a NULL check, while `ngx_http_modsecurity_process_intervention()` does check it.
- **`docs/phase4-handling.de.md`**: I'd rather not carry a 251-line German translation in-tree - it will drift from the English version and none of the maintainers can review changes to it. Could you drop it and keep the English doc as the single source?
- **`tests/modsecurity-phase4-content-types.t`**: the plan still asserts `"content_type":"application\/json"` twice (checks 4 and 9) - leftover from the Copilot comment about the duplicated alternation.

I haven't built this against libmodsecurity yet, so the C99 mid-block declarations and the `%ui`/`%i` format usage still need a real compile on a couple of toolchains; I'll verify locally too.

To be clear, I like the direction - making late `phase:4` behavior explicit instead of pretending `deny` works after headers are sent is the right call, and the documentation of the limitation is valuable. It's points 1 and 2 that I'd like to resolve first.

I used to make this review with help of Claude Opus 5.

Thank you again!

Easton97-Jens:
Okay, I'll take a look at it; I've already fixed the config issue in the repo.
https://github.com/Easton97-Jens/ModSecurity-conector
I'm going to install an on/off switch, remove a tiny bit, and take out the mime as well.

PR #378, opened 2026-05-13
fix: free intervention.url on redirect path and check ngx_list_push r…

eilandert:
…eturn

When ModSecurity fires a redirect intervention, intervention.url is allocated by msc_intervention() and must be freed by the caller. Previously it was leaked on every redirect, causing unbounded memory growth on busy servers using redirect-based rules.

Additionally, the ngx_list_push() call was not checked for NULL before dereferencing the returned pointer, which would cause a worker crash under out-of-memory conditions.

Fixes:
- Add free(intervention.url) before returning the redirect status code
- Add NULL check for ngx_list_push() return; free url and return 500 on allocation failure

### DIFF
--- a/src/ngx_http_modsecurity_module.c
+++ b/src/ngx_http_modsecurity_module.c
@@ -208,6 +208,10 @@ ngx_http_modsecurity_process_intervention (Transaction *transaction, ngx_http_re
 
         ngx_table_elt_t *location = NULL;
         location = ngx_list_push(&r->headers_out.headers);
+        if (location == NULL) {
+            free(intervention.url);
+            return NGX_HTTP_INTERNAL_SERVER_ERROR;
+        }
         ngx_str_set(&location->key, "Location");
         location->value = a;
         r->headers_out.location = location;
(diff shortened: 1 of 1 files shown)


airween:
@eilandert,

thanks for the PR - do you think it would be possible to add some tests to check this feature?

tomsommer:
I can confirm the leak this PR targets: under valgrind (nginx master, libmodsecurity 3.0.14, `--leak-check=full`) the `strdup` from `Transaction::intervention()` on the redirect path shows up as 680 bytes / 20 blocks after 20 phase 1 redirects and 6,800 bytes / 200 blocks after 200, so it grows per request.

The fix as written turns it into a use-after-free, though. A few lines above the new `free()`, `a.data` is set to `intervention.url` and `location->value = a;` stores that pointer in the response headers. The `Location` header is only serialized later by `ngx_http_header_filter`, after this function has returned, so freeing the buffer here leaves the header value pointing at released memory. The URL needs to be copied into `r->pool` first, for example:

```c
a.len = ngx_strlen(intervention.url);
a.data = ngx_pnalloc(r->pool, a.len);
if (a.data == NULL) {
    free(intervention.url);
    return NGX_HTTP_INTERNAL_SERVER_ERROR;
}
ngx_memcpy(a.data, intervention.url, a.len);
free(intervention.url);
intervention.url = NULL;
```

The `location == NULL` branch you added is fine as is.

For reference, #397 fixes the other leak of the same string, on the `r->header_sent` early return, and its test asserts the exact `Location` value on a phase 1 redirect, which would catch a dangling header. Found during a valgrind audit of the connector done with Claude Fable 5.1.

PR #379, opened 2026-05-13
fix: free intervention strings on mcf NULL early return path

eilandert:
When msc_intervention() succeeds and fills intervention.log and/or intervention.url, then ngx_http_get_module_loc_conf() returns NULL (unexpected misconfiguration), the function returned immediately with NGX_HTTP_INTERNAL_SERVER_ERROR without freeing either string.

Free both intervention.log and intervention.url before the early return to prevent the leak on this error path.


airween:
Hi @eilandert,

thanks for this PR too - here my question is the same: is it possible to add some tests?

PR #381, opened 2026-05-13
fix: document request body plaintext disk spooling and mitigation steps

eilandert:
request_body_in_persistent_file unconditionally spools all client request bodies to disk as plaintext temporary files. This bypasses client_body_buffer_size, meaning even small bodies end up on disk.

In environments handling sensitive POST data (authentication, payment, health records), this creates plaintext files readable by any process with access to client_body_temp_path, which defaults to a world- accessible directory under the nginx prefix.

Add a security comment at the point of the flag assignment documenting:
- The nature of the exposure (plaintext temp files regardless of buffer)
- The recommended mitigation: client_body_temp_path to a tmpfs mount such as /dev/shm/nginx_body with 0700 permissions
- The recommended size limits to constrain disk usage

A proper fix would only spool to disk when body size exceeds client_body_buffer_size and use the configured temp path with restrictive permissions. This comment serves as a clear callout for operators reviewing the source and for future refactoring.

### DIFF
--- a/src/ngx_http_modsecurity_access.c
+++ b/src/ngx_http_modsecurity_access.c
@@ -343,6 +343,23 @@ ngx_http_modsecurity_access_handler(ngx_http_request_t *r)
          * r->request_body_in_single_buf = 1;
          */
         r->request_body_in_single_buf = 1;
+        /*
+         * SECURITY NOTE: request_body_in_persistent_file forces all request
+         * bodies to be written to disk as plaintext temporary files regardless
+         * of the client_body_buffer_size setting. Sensitive data in POST bodies
+         * (passwords, tokens, PII) will be written to the directory specified
+         * by client_body_temp_path (default: a prefix/client_body_temp).
+         *
+         * To reduce exposure, set client_body_temp_path to a tmpfs/ramfs mount
+         * in your nginx configuration, for example:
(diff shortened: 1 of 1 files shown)


tomsommer:
Confirming the behaviour this comment documents, with two corrections to the wording that I verified against the nginx source:

- Bodies are not spooled to disk unconditionally. `request_body_in_persistent_file` only changes what happens to the temp file once nginx decides to create one; the decision itself is still made by `client_body_buffer_size` (`ngx_http_request_body.c`): bodies that fit the buffer stay in memory, larger ones are written to `client_body_temp_path`. So "bypasses client_body_buffer_size" and "even small bodies end up on disk" are not accurate.
- The flag is required rather than optional: `ngx_create_temp_file()` unlinks a non-persistent temp file immediately after creating it, and the connector hands the file to libmodsecurity by name (`msc_request_body_from_file`), which reopens it and reads it back into memory in full. That read-back is also why the spill saves no memory in practice.

The mitigation advice (tmpfs-backed `client_body_temp_path`, sizing `client_body_buffer_size`) is sound. #393 documents the same behaviour for operators in the README, worded so that it agrees with the corrected version of this comment. Found during a performance review of the connector done with Claude Fable 5.1.

PR #391, opened 2026-09-19
Add modsecurity_request_body directive to skip request body buffering (supersedes #356)

### LINKED ISSUE #343: Add an Option to Skip Body Inspections

tomsommer:
## what

- New directive `modsecurity_request_body on | off` (http, server, location; default `on`, inherited per location).
- With `off` the connector no longer calls `ngx_http_read_client_request_body()` in the ACCESS phase, no longer sets the `request_body_in_*` buffering flags, and feeds no body to libmodsecurity. The REQUEST_BODY phase (`msc_process_request_body`) still runs and its intervention is still honoured, so phase 2 rules on `ARGS`, `REQUEST_HEADERS`, etc. keep working; `REQUEST_BODY`, `ARGS_POST` and `FILES` are simply empty.
- README section for the directive; the outdated "adds four new directives" sentence becomes count-free.
- New test `tests/modsecurity-request-body-directive.t` (9 assertions): default still blocks a bad body, `off` lets it through while `ARGS` rules still block, `off` + `proxy_request_buffering off` delivers the full body to the upstream, and `on` in a nested location overrides an inherited `off`.

## why

- The connector reads and buffers the **whole** request body before the content handler runs, so `proxy_request_buffering off` is ineffective in every location where `modsecurity` is on, and large uploads are held in memory or spooled to a temp file before being proxied.
- libmodsecurity copies the body in `Transaction::appendRequestBody()` regardless of `SecRequestBodyAccess` (only `processRequestBody()` checks it), so turning body access off in the rules does not avoid the buffering cost. The connector cannot query that setting through the C API, hence an nginx-level directive.
- Typical uses: large or encrypted uploads, streaming APIs, and locations that already run with `SecRequestBodyAccess Off`.

## references

- Closes #343 (Add an Option to Skip Body Inspections).
- Supersedes #356: that PR returns from the access handler before the transaction is created, so `off` disabled the whole WAF for the location rather than only the body, and its new conf fields were never merged, so the setting did not inherit (both points raised in its review). This PR keeps the transaction, still runs phase 2, merges the flag with `ngx_conf_merge_value`, and ships the tests the reviewers asked for.
- Related: #381 (temp-file spooling comment).

---

Origin: this change comes from a performance review of the connector done with Claude Fable 5.1 (Anthropic). The patch and its test were verified by building the module against nginx master with libmodsecurity 3.0.14 (PCRE2) and, with upstream CI's flags (`--without-pcre2 --with-http_v2_module --with-http_auth_request_module`), against libmodsecurity 3.0.9 (PCRE1), then running the full `tests/modsecurity*.t` suite in both builds (16 files, 260 tests, all passing).

## Summary by CodeRabbit

- **New Features**
  - Added the `modsecurity_request_body` directive to control request-body inspection at the main, server, or location level.
  - Request-body inspection is enabled by default; disabling it allows bodies to stream to upstream services while preserving other request-processing checks.
  - Added documentation covering defaults, configuration contexts, behavior, and upload-location usage.

- **Tests**
  - Added coverage for default, inherited, overridden, disabled, streaming, proxy-buffered, and GET request behavior.


airween:
@tomsommer: thanks for the PR.

I see so many advantages here against to use library's `SecRequestBodyAccess` [directive](https://github.com/owasp-modsecurity/ModSecurity/wiki/Reference-Manual-(v3.x)#user-content-SecRequestBodyAccess):
* can be used in several contexts: _http, server, location; default on, inherited per location_
* avoid to pass the whole body to the library

and I see only one disadvantage: the admin can be confused with the library's directive.

Anyway, I think this is a good direction and want to merge this (wait for @thekief and @HanadaLee's opinion here).

Also we have to investigate why the Windows tests are failed... (And probably some previous PR's will be merged, so I ask for your patience.

Thank you again.

tomsommer:
Thanks for looking at it so quickly.

**On the naming.** That is a fair concern. The way I would describe the split is: `SecRequestBodyAccess` tells the library whether to *evaluate* the body, this directive tells the connector whether to *collect and hand over* the body at all. They are not redundant, because libmodsecurity copies the body in `Transaction::appendRequestBody()` before `processRequestBody()` ever looks at `SecRequestBodyAccess`, so the copy happens either way and the connector cannot see that setting through the C API. The README section says exactly that, to keep the two apart for the admin. If you would still prefer a name that cannot be mistaken for the library directive, I am happy to rename; `modsecurity_request_body_buffering` would be the honest description of what it does. Whatever is chosen should apply to #395's `modsecurity_response_body` as well, so the pair stays symmetric.

**On the test.** Copilot made a good point that the original assertion only checked the final byte count, which a still-buffering implementation would also satisfy. I have pushed a real streaming regression test: the client sends 6 of 13 body bytes and stops, and the upstream answers with what it has already received. With the directive off and `proxy_request_buffering off` the upstream sees the partial body (`EARLY LEN=6`); with the directive on it sees nothing until the client completes the body, and the completed body is still inspected and blocked. The negative case fails on `master` and on a build with the directive forced on, so it pins the actual behaviour rather than the byte count.

**On the Windows jobs.** They are not caused by these PRs. The three `windows-nginx-1.28/1.29/1.30` jobs from `Quality Assurance new` fail on every open PR right now, including #384, #385, #386, #389 and the documentation-only #393, and the same workflow last succeeded on `master` in May 2026.

The failure is in dependency provisioning, before anything of ours is compiled:

```
yajl/2.1.0: Running CMake.configure()
yajl/2.1.0: RUN: cmake -G "Visual Studio 18 2026" ...
CMake Error at CMakeLists.txt:17 (cmake_policy):
  Policy CMP0026 may not be set to OLD behavior because this version of CMake
  no longer supports it.
...
ERROR: yajl/2.1.0: Error in build() method, line 53
```

Everything after that is fallout (`The source directory .../nginx/objs/lib/ModSecurity/build does not appear to contain CMakeLists.txt`).

Two things changed on the runner side and together they produce this:

1. The image now ships Visual Studio 18 / MSVC 19.51 (toolset v145), so Conan Center has no prebuilt `yajl/2.1.0` binary for that ABI and falls back to building it from source.
2. Conan provisions `cmake/4.4.3` for that build, and CMake 4 removed `OLD` support for `CMP0026`, which yajl 2.1.0's `CMakeLists.txt` still sets.

So the fix belongs in the workflow, not in the PRs. The options I can see are pinning the CMake that Conan uses for the dependency build (a `tool_requires` on a 3.x CMake in the profile), taking a newer `yajl` recipe revision that no longer sets the old policy, or dropping yajl from the Windows dependency set if it is not actually needed there. Happy to open a separate PR for whichever you prefer, so it does not get mixed into these changes.

No rush at all on merging, and it makes sense to land the earlier PRs first. If you would like any of these rebased once those go in, just say the word.

PR #395, opened 2026-09-19
Add modsecurity_response_body directive to skip response body buffering (closes #206)

### LINKED ISSUE #206: [Perf] Don't read full response into mem if not needed NGINX

tomsommer:
## what

- New directive `modsecurity_response_body on | off` (http, server, location; default `on`, inherited per location).
- With `off`:
  - the header filter no longer sets `r->filter_need_in_memory`, so nginx no longer reads file-backed response buffers (static files, proxied responses buffered to disk) into memory and copies them through the output chain before sending them (with `sendfile` enabled they are transmitted straight from the file, as they would be without the module);
  - the body filter no longer calls `msc_append_response_body()` or the per-buffer intervention check;
  - the RESPONSE_BODY phase is still invoked on the last buffer, so libmodsecurity's own `SecResponseBodyAccess` / `SecResponseBodyMimeType` logic still decides whether phase 4 rules run; `RESPONSE_BODY` is empty and `RESPONSE_CONTENT_LENGTH` is 0.
- README section for the directive; the outdated "adds four new directives" sentence becomes count-free.
- New test `tests/modsecurity-response-body-directive.t` (13 assertions): audit-log evidence that a `RESPONSE_BODY` rule matches by default and not with `off` while a status-based phase 4 rule still runs, the body is delivered intact, inheritance into a nested location works, and, on `--with-debug` builds, the copy filter's "write new buf t:1 f:1" record appears only for the default location.

## why

- Closes #206. The module unconditionally forces every response body into memory, in every location where `modsecurity` is on, regardless of `SecResponseBodyAccess`. The FIXME next to that line has said so since 2017.
- libmodsecurity's `appendResponseBody()` only filters on the MIME-type list, not on `SecResponseBodyAccess` (that is checked in `processResponseBody()`), so turning body access off in the rules does not avoid the copy, and the connector cannot query that setting through the C API. An nginx-level directive is the only lever.
- Response buffers are handed to the client as soon as they have been fed to ModSecurity, so phase 4 can only ever affect the last chunk of a streamed response; sites that do not rely on it pay the full copy for nothing.

## references

- Closes #206.
- Overlaps #356's `modsecurity_skip_resp_body_filter`: that variant skips the body filter but leaves `filter_need_in_memory` set (so the copy into memory remains) and never runs the RESPONSE_BODY phase; this PR does both and adds tests. Naming follows #391 (`modsecurity_request_body`).
- Leaves the `return ret;` on the last-buffer intervention path in the body filter untouched because #384 owns that block.

---

Origin: this change comes from a performance review of the connector done with Claude Fable 5.1 (Anthropic). The patch and its test were verified by building the module against nginx master with libmodsecurity 3.0.14 (PCRE2) and, with upstream CI's flags (`--without-pcre2 --with-http_v2_module --with-http_auth_request_module`), against libmodsecurity 3.0.9 (PCRE1), then running the full `tests/modsecurity*.t` suite in both builds (16 files, 264 tests, all passing).

## Summary by CodeRabbit

* **New Features**
  * Added the `modsecurity_response_body` directive to control whether response bodies are inspected by ModSecurity.
  * Response-body inspection is enabled by default and can be disabled at the location, server, or main configuration level.
  * Disabled inspection avoids buffering response data while preserving subsequent request processing.

* **Documentation**
  * Added usage guidance for configuring response-body inspection.

* **Tests**
  * Added coverage for default, disabled, and inherited configuration behavior.


airween:
Hi @tomsommer,

thanks for this too.

So it's similar to #391 but for the response body, right? And this bypasses the engine's `SecResponseBodyAccess` [directive](https://github.com/owasp-modsecurity/ModSecurity/wiki/Reference-Manual-(v3.x)#secresponsebodyaccess)?

Thank you for this, I have same opinion here like in case of the other PR. I ask for your patience here too.

tomsommer:
Hi @airween, thanks for looking at this one too.

Yes, it is the response-side counterpart of #391, but I would not describe it as bypassing `SecResponseBodyAccess`, because it cannot make the engine inspect anything the admin turned off. The two settings act at different points:

* `SecResponseBodyAccess` decides whether **libmodsecurity evaluates** the response body (checked in `Transaction::processResponseBody()`).
* `modsecurity_response_body` decides whether **the connector collects and hands over** the body at all.

With the directive at its default `on`, behaviour is exactly as it is today: every buffer is passed to the library and `SecResponseBodyAccess` together with `SecResponseBodyMimeType` decides what happens to it. With `off`, the connector hands nothing over, so `RESPONSE_BODY` is empty and `RESPONSE_CONTENT_LENGTH` is 0 no matter how the engine is configured. The connector still calls the response-body phase on the last buffer, so phase 4 rules that do not look at the body (for example on `RESPONSE_STATUS`) keep working, subject as always to the engine's own decision whether to run phase 4 at all.

The reason a connector-side switch is needed even when an admin has already set `SecResponseBodyAccess Off` is that the cost is paid before the engine's setting is consulted, in two places:

1. `Transaction::appendResponseBody()` filters only on `SecResponseBodyMimeType`; it does not look at `SecResponseBodyAccess`, which is checked later in `processResponseBody()`. So the copy into the library's buffer happens either way, and the connector cannot query that setting through the C API.
2. More importantly, the header filter sets `r->filter_need_in_memory = 1` unconditionally. That makes nginx read file-backed responses (static files, or proxied responses buffered to disk) into memory and copy them through the output chain before sending them, even when no rule will ever look at them. `sendfile` still performs the transmission, so the cost is the extra read and copy rather than the loss of `sendfile`, but it is paid on every response in a location where `modsecurity` is on.

That second point is what #206 has been asking for since 2020, and it is the part that is not reachable from the rules language at all. It also makes this PR slightly more than a mirror image of #391: the request side is only about not reading and buffering the body, the response side additionally stops forcing the body into memory.

On naming, the same offer as in #391 applies: if `modsecurity_response_body` reads too close to `SecResponseBodyAccess`, I am happy to rename both directives to something like `modsecurity_request_body_buffering` and `modsecurity_response_body_buffering`, which describes what they actually control. I would keep whatever is chosen symmetric across the two PRs.

No rush from my side, and it makes sense to land the earlier PRs first. Happy to rebase whenever that happens.

